PR #20, opened 2022-03-28
News translations fixes

SebaKas21:


be/src/Infrastructure/Localization/Foundation.Localization/Resources/GlobalResources.de.resx
```diff
@@ -445,9 +445,9 @@
     <value>Korb</value>
   </data>
   <data name="NewsCategories" xml:space="preserve">
-    <value>News Kategorien</value>
+    <value>Aktuell Kategorien</value>
```
dawiddworak88, line 448: There is too much "Aktuell" :D I'll help you with this one, don't worry :)

PR #21, opened 2022-03-28
Reverted categories changes

SebaKas21:


be/src/Project/Web/Seller/Seller.Web/Areas/Products/ModelBuilders/CategoryFormModelBuilder.cs
```diff
@@ -70,7 +70,7 @@ public async Task<CategoryFormViewModel> BuildModelAsync(ComponentModelBase comp
             var parentCategories = await this.categoriesRepository.GetAllCategoriesAsync(
                 componentModel.Token,
                 componentModel.Language,
-                null, $"{nameof(Category.Level)}");
+                null, $"{nameof(Category.Level)},{nameof(Category.Name)}");
```
dawiddworak88, line 73: Hi Sebastian, this change will be much more complicated. I'll show and tell you why. We'll discuss it on Wednesday.

PR #22, opened 2022-03-29
Back buttons for each form

SebaKas21:


dawiddworak88 (commented):
Hi @SebaKas21 , almost right. The idea is to have both buttons always visible next to each other. The "Back to..." button should be styled as secondary.

image

dawiddworak88 (changes requested):
Hi @SebaKas21 , it would be great you go through all these forms and check how they look on desktop and mobile. Check in material-ui docs how to place two buttons next to each other. I'd like to ask you to style the "Back to List" buttons as a secondary buttons (they should differ from the primary-styled "Save" button):

image

SebaKas21:
@dawiddworak88 

Buttons are styled on secondary color.
Secondary color value is set like primary color

obraz

Buttons looks good at mobile version

obraz

be/src/Infrastructure/Localization/Foundation.Localization/Resources/ProductResources.en.resx
```diff
@@ -162,6 +162,18 @@
...
     <value>Published</value>
   </data>
+  <data name="NavigateToAttributesLabel" xml:space="preserve">
+    <value>Back to attributes</value>
+  </data>
+  <data name="NavigateToCategoriesLabel" xml:space="preserve">
+    <value>Back to categories</value>
+  </data>
+  <data name="NavigateToProductAttributeLabel" xml:space="preserve">
+    <value>Back to attribute</value>
```
dawiddworak88: **suggestion:** Keep the convention:  "Back to attributes"

be/src/Infrastructure/Localization/Foundation.Localization/Resources/ProductResources.en.resx
```diff
@@ -162,6 +162,18 @@
...
+  </data>
+  <data name="NavigateToProductsLabel" xml:space="preserve">
+    <value>Back to products lists</value>
```
dawiddworak88: **suggestion:** Keep the convention. It should be "Back to products"

be/src/Infrastructure/Localization/Foundation.Localization/Resources/ProductResources.pl.resx
```diff
@@ -162,6 +162,18 @@
...
+    <value>Powróć do atrybutów</value>
+  </data>
+  <data name="NavigateToCategoriesLabel" xml:space="preserve">
+    <value>Powróć do kategorii</value>
+  </data>
+  <data name="NavigateToProductAttributeLabel" xml:space="preserve">
+    <value>Powróć do atrybutu</value>
+  </data>
+  <data name="NavigateToProductsLabel" xml:space="preserve">
+    <value>Powróc do listy produktów</value>
```
dawiddworak88: **suggestion:** Fix typos and keep the convention. It should be "Powróć do produktów"

fe/src/project/Seller.Portal/areas/Inventory/components/InventoryAddForm/InventoryAddForm.js
```diff
@@ -72,7 +72,7 @@ const InventoryAddForm = (props) => {
                 return res.json().then(jsonRes => {
                     if (res.ok) {
                         toast.success(jsonRes.message);
-                        setShowBackToInventoryListButton(true);
+                        setDisableSaveButton(true);
```
dawiddworak88, line 75: **suggestion:** We should get rid of it. We have already the disable property. It's not present in many other places like ClientsForm above

dawiddworak88 (changes requested):
Hi @SebaKas21 , please go this PR one line after the other to fix the typos and conventions. I gave you just a few examples of what to look for. It would be also great to use the exisitng 'disable' rather than introducing new properties.

fe/src/project/Seller.Portal/areas/Inventory/components/InventoryAddForm/InventoryAddForm.js
SebaKas21: @dawiddworak88 

I know that I can operate on the id in useForm, but think about the PR related to displaying the id, even after saving I have to use this state because I will not display information about the id

dawiddworak88 (changes requested):
Hi @SebaKas21,

this one needs very careful testing, actually one entity after the other. I was going through and found out that it's not possible to save news categories:

image

Cheers,

dawiddworak88 (approved):

PR #24, opened 2022-03-30
Header responsive improvements

SebaKas21:


dawiddworak88 (changes requested):
Hi @SebaKas21 , it looks a bit too tight on mobile and there is no padding at all on the search button:

image

dawiddworak88 (changes requested):
Hi @SebaKas21 , we need to fix two more things on mobile here:

1.  There should be some air between the search bar and the button, a bit of space
2. There is some strange arrow upon search

image

dawiddworak88 (approved):
Let's take the updated header :)

PR #30, opened 2022-04-04
Updated news translation

SebaKas21:


dawiddworak88 (approved):
Short but effective, approved :)

PR #31, opened 2022-04-05
Space for file download button

SebaKas21:


dawiddworak88 (changes requested):
HI @SebaKas21 , sorry I made wrong screenshot. Couldn't we implement it without using !important statements?

SebaKas21:
@dawiddworak88 

I decided to create a class, because I had no idea, I used bulma but that I would not give me any space

dawiddworak88 (changes requested):
Hi @SebaKas21 , unfortunately it still looks like on the picture below. We need to fix it differently, without using !important statements. They are bad practice and should be avoided. 

image

fe/src/project/AspNetCore/shared/components/Files/Files.scss
```diff
@@ -0,0 +1,10 @@
+.file-tooltip {
```
dawiddworak88: **suggestion:** try to stick to BEM convention: http://getbem.com/naming/ It should be .files__tooltip

fe/src/project/AspNetCore/shared/components/Files/Files.scss
```diff
@@ -0,0 +1,10 @@
...
+    margin-left: 0.5rem !important;
```
dawiddworak88: **suggestion:** It would be great to get rid of !important statements.

dawiddworak88 (changes requested):

fe/src/project/AspNetCore/shared/components/Files/Files.scss
SebaKas21: Thanks

dawiddworak88 (approved):
🥇

PR #32, opened 2022-04-05
Self-Service Password Reset

SebaKas21:


dawiddworak88 (changes requested):
Hi @SebaKas21 ,

1. Please use the standard bulma "button is-ghost" styling without adjusting size. There is at the moment too much space between the reset link and the button.
2. The text on the button should be formed as call to action "Resetuj hasło" instead of "Resetowanie hasła".

SebaKas21:
@dawiddworak88 

I think that button look bad, what you think?
obraz

dawiddworak88 (changes requested):
1. Still 'Resetowanie hasła' instead of CTA 'Zresetuj hasło'
2. is-ghost styling doesn't work and looks completely different from what's in the docs. It could have been introduced in newer versions of Bulma. Please check and use is-text if that's the case. Please fix the margin-top, so that the distance between the reset link and log in buttons are smaller
3. Exception is thrown when false non-existing e-mail is sent. The right behavior for security reason is to go successfully to the next step. Change the success message: "The reset password link has been sent if your account exists in our platform. Please check your e-mail box for further instructions."

image

dawiddworak88 (approved):
Well done, thank you @SebaKas21

PR #36, opened 2022-04-11
Client phone number

SebaKas21:


fe/src/project/Seller.Portal/areas/Clients/components/ClientForm/ClientForm.js
```diff
@@ -173,6 +181,21 @@ function ClientForm(props) {
                                 )}
                             </FormControl>
                         </div>
+                        <div className="field">
```
dawiddworak88, line 179: @SebaKas21 please use the regular TextField here. I don't want to hardcode regions like "europe" here.

fe/src/project/Seller.Portal/areas/Clients/pages/ClientPage/ClientPage.js
```diff
@@ -18,7 +18,7 @@ function ClientPage(props) {
         <div>
           <Header {...props.header}></Header>
           <MenuTiles {...props.menuTiles} />
-          <ClientForm {...props.clientForm} />
+          <ClientForm {...props.clientForm} locale={props.locale} />
```
dawiddworak88: @SebaKas21 , when we use TextField instead of ReactPhoneInput it can be safely removed. I'd give flexibility here instead of restricting it too much.

dawiddworak88 (changes requested):

fe/src/project/Seller.Portal/areas/Clients/components/ClientForm/ClientForm.js
SebaKas21: @dawiddworak88 ReactPhoneInput should be deleted?

dawiddworak88 (approved):
🚀

PR #38, opened 2022-04-12
Inventory improvements

SebaKas21:


dawiddworak88:
Why are you trying to remove it completely? We should never allow to let sellers see warehouses of other sellers. We'll discuss it.

SebaKas21:
Problem is when client want get list of inventory items, he received: nullable object error. Seller Id is coming from jwt

PR #39, opened 2022-04-13
Order cancellation button

SebaKas21:
I added buttons below other components, please look and give me feedback :)


be/src/Project/Web/Seller/Seller.Web/Areas/Orders/ApiControllers/OrderStatusApiController.cs
```diff
@@ -26,6 +27,17 @@ public class OrderStatusApiController : BaseApiController
             this.orderLocalizer = orderLocalizer;
         }
 
+        [HttpPost]
+        public async Task<IActionResult> Cancel([FromBody] UpdateOrderStatusRequestModel model)
+        {
+            var token = await HttpContext.GetTokenAsync(ApiExtensionsConstants.TokenName);
+            var language = CultureInfo.CurrentUICulture.Name;
+
+            var orderStatusId = await this.ordersRepository.SaveOrderStatusAsync(token, language, model.OrderId, OrdersConstants.OrderStatuses.CancelId);
```
dawiddworak88, line 42: **suggestion:** Please add check if the order state is in new. If not, throw a custom exception with a meaningful message that only newly created orders can be cancelled.

fe/src/project/Seller.Portal/areas/Orders/components/EditOrderForm/EditOrderForm.js
```diff
@@ -111,7 +146,17 @@ function EditOrderForm(props) {
                             <a href={props.clientUrl}>{props.clientName}</a>
                         </div>
                     </div>
-                    
+                    <div className="mt-5">
```
dawiddworak88: suggestion: Please add logic to show the button only for orders with an order state new. It makes no sense to provoke users with this button for orders that are already in progress, closed, etc.

dawiddworak88 (changes requested):

dawiddworak88:
@SebaKas21 , could you merge develop to this feature branch as well?

SebaKas21:
@dawiddworak88  Can we merge?

dawiddworak88 (changes requested):
Hi @SebaKas21 , we need to extend it. The cancel button should be also added to each order item to add a possibility to cancel every order item. The cancellation button on the order should display a modal window with a warning that it will cancel all order items. Thereafter the order items should be cancelled and the order at the end.

dawiddworak88 (approved):
Merci 👍

PR #43, opened 2022-04-19
EAN on the Sidebar

SebaKas21:


fe/src/project/AspNetCore/shared/components/Sidebar/Sidebar.scss
```diff
@@ -107,6 +107,7 @@
         flex-direction: column;
 
         .stock-details {
+            margin-top: 10px;
```
dawiddworak88: @SebaKas21 , could we make it in rem units?

dawiddworak88 (approved):

PR #45, opened 2022-04-20
Clients updating without account

SebaKas21:
Hi @dawiddworak88 

I added improvements for client update request, i removed redundant request /users/{id} it wasn't use anywhere, and created request with email query (i think that will be used more, because email for us is like primary key)


dawiddworak88 (approved):
🥇

PR #46, opened 2022-04-21
Draggable menu tiles

SebaKas21:
I moved this to another PR.

I know i must experiment at this one


dawiddworak88:
@SebaKas21: Let's go for alternative solution with a simple button to avoid issues

PR #48, opened 2022-04-21
Fixed possibility basket duplication

SebaKas21:


dawiddworak88 (approved):
Thank you for spotting it, well done @SebaKas21

PR #51, opened 2022-04-25
Smooth tiles scrolling

SebaKas21:


dawiddworak88:
Hi @SebaKas21, I'll close this PR because the implementation is very complicated, looks suboptimal and doesn't support screen readers (arrows left and right). Could you please try to implement it using react-slick? https://github.com/akiran/react-slick 

An example of how it should work like you'll find in the Responsive Display section here: http://kenwheeler.github.io/slick/

PR #62, opened 2022-05-06
Environment update to node v18

SebaKas21:


SebaKas21:
@dawiddworak88  

I think, now problem is resolved.

dawiddworak88 (approved):
Let's go after it 🚀

dawiddworak88:
Hi @SebaKas21 , could you pull from develop, nvm use 18.1.0, npm install --legacy-peer-deps, npm run build-fe, go through the pages one after the other and fix problems like these:

image

PR #70, opened 2022-05-11
Fixed ordering unpublished products

SebaKas21:


dawiddworak88:
Hi @SebaKas21 , this solution is incorrect for two reasons:

1. It still allows to place orders for unpublished products directly via API
2. It performs filtering once the list of products is already retrieved from the repository. It causes a lot of redundant data being transferred through the wire. 

I'll close this PR for now and we'll discuss the right solution.

PR #71, opened 2022-05-11
Deleting products from inventory when product is deleted

SebaKas21:


be/src/Project/Services/Catalog/Catalog.Api/Services/Products/ProductsService.cs
```diff
@@ -285,6 +285,16 @@ public async Task DeleteAsync(DeleteProductServiceModel model)
...
 
+            var message = new DeletedProductIntegrationEvent
+            {
+                ProductId = model.Id,
+                Language = model.Language,
+                Username = model.Username,
+                OrganisationId = model.OrganisationId
+            };
+
+            this.eventBus.Publish(message);
```
dawiddworak88, line 298: **suggestion:** it should be after await this.catalogContext.SaveChangesAsync(); because if saved to the database fails it makes no sense to publish the event.

dawiddworak88 (changes requested):

be/src/Project/Services/Catalog/Catalog.Api/Services/Products/ProductsService.cs
SebaKas21: @dawiddworak88  

You've right

dawiddworak88 (approved):

PR #74, opened 2022-05-16
Account kiosk for clients

SebaKas21:
Hi @dawiddworak88,

Check if such inputs are okay, or add or take something :)
There is a form to be improved, because the Radio has a problem and they are going away and the code quality correction, which I will of course also do :D

Radio problem (i will fix at the morning tomorrow)
obraz


SebaKas21:
My question for you @dawiddworak88, where we want redirect after successfully apply? To login page?

dawiddworak88 (approved):
Let's take it :)

PR #77, opened 2022-05-18
Double quantities instead of int

SebaKas21:


dawiddworak88:
@SebaKas21 , could you please check whether inventory sync and decreasing quantities work properly after merge?

dawiddworak88 (approved):
Let's merge it

PR #78, opened 2022-05-18
Custom order details

SebaKas21:
Removed redundant rows parametr, because when order has more rows then 3, textarea is displaying only 3 rows


dawiddworak88 (approved):
OK.

PR #90, opened 2022-06-01
Improvements for setting password after reset

SebaKas21:
Hi @dawiddworak88 

I think this PR should solve the password reset problem, if not then I would replace:
```c#
if (existingUser.EmailConfirmed is false)
```
on:

```c#
if (existingUser.EmailConfirmed is false && (existingUser.VerifyExpirationDate >= DateTime.UtcNow))
```


SebaKas21:
@dawiddworak88  I find next problem with redirect after successful password reset

SebaKas21:
@dawiddworak88 Redirect should be now okey

dawiddworak88 (approved):

PR #92, opened 2022-06-03
Added list of clients applications

SebaKas21:


be/src/Project/Services/Client/Client.Api/IntegrationEventsHandlers/ClientApplicationIntegrationEventHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            this.clientsApplicationsService = clientsApplicationsService;
+        }
+
+        /// <summary>
+        /// Integration event handler which saving
+        /// </summary>
+        /// <param name="event">
+        /// Integration event message which is sent by the
+        /// basket.api once it has successfully process the
```
dawiddworak88: suggestion: too much copy-paste? ;)

be/src/Project/Services/Client/Client.Api/Program.cs
```diff
@@ -67,6 +67,8 @@
 
 builder.Services.RegisterDatabaseDependencies(builder.Configuration);
 
+builder.Services.RegisterEventBus(builder.Configuration);
```
dawiddworak88: **question:** why do we need this?

dawiddworak88 (changes requested):

be/src/Project/Services/Client/Client.Api/Program.cs
SebaKas21: Hi @dawiddworak88 ,

We need it because we previously used the endpoint created in the identity api, not to confuse it, I send it by queue to client.api

dawiddworak88 (changes requested):
Hi @SebaKas21 ,

could you please as discussed remove the RegisterAsync from Identity.Api, move it to Buyer.Web and get rid of all the EventBus references?

dawiddworak88 (approved):

PR #95, opened 2022-06-13
Confirmation dialog delete name property

SebaKas21:


be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ModelBuilders/ClientAccountManagersPageCatalogModelBuilder.cs
```diff
@@ -51,6 +51,7 @@ public async Task<CatalogViewModel<ClientAccountManager>> BuildModelAsync(Compon
             viewModel.SearchApiUrl = this.linkGenerator.GetPathByAction("Get", "ClientAccountManagersApi", new { Area = "Clients", culture = CultureInfo.CurrentUICulture.Name });
 
             viewModel.OrderBy = $"{nameof(ClientAccountManager.CreatedDate)} desc";
+            viewModel.ConfirmationDialogDeleteNameProperty = nameof(ClientAccountManager.FirstName).ToCamelCase();
```
dawiddworak88: **suggestion:** it should be an IEnumerable<string>, so that a list of properties can be passed. It's a good example that for account manager it should be FirstName and LastName of Account Manager displayed in the delete dialog :)

dawiddworak88 (changes requested):

dawiddworak88 (approved):

PR #96, opened 2022-06-13
Renamed Client Managers to Client Account Managers

SebaKas21:


be/src/Infrastructure/Localization/Foundation.Localization/Resources/ClientResources.en.resx
```diff
@@ -136,7 +136,7 @@
     <value>A customer with this e-mail address already exists.</value>
   </data>
   <data name="ClientManagers" xml:space="preserve">
-    <value>Client  account managers</value>
+    <value>Client account managers</value>
```
dawiddworak88, line 139: **suggestion:** there are probably some more labels to have a look at:

image

dawiddworak88 (changes requested):

SebaKas21:
Hi @dawiddworak88 

The only place I haven't changed is the menu.

obraz

dawiddworak88 (approved):
Perfect 👍

PR #97, opened 2022-06-14
Redirection after password reset

SebaKas21:
Hi @dawiddworak88 

Client mobel.lt writing to it.developers with password reset issue,

In our logs i find it problem and testing and my solution is add variable to deployment
Cheers

PS: ReturnUrl was always null


dawiddworak88 (approved):
Perfect, thank you for taking care of it!

PR #98, opened 2022-06-20
Renemed Medias to MediaItems

SebaKas21:


be/src/Project/Web/Seller/Seller.Web/Areas/Media/ApiRequestModels/UploadMediaItemRequestModel.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Seller.Web.Areas.Media.ApiRequestModels
 {
-    public class UploadMediaRequestModel : RequestModelBase
+    public class UploadMediaItemRequestModel : RequestModelBase
```
dawiddworak88: **suggestion:** it's too much of a change. We should only get rid of the 'Medias'. I'd leave everything else as it was.

dawiddworak88 (changes requested):

be/src/Project/Web/Seller/Seller.Web/Areas/Media/ApiRequestModels/UploadMediaItemRequestModel.cs
SebaKas21: Okey

SebaKas21:
Hi @dawiddworak88 

Now will be okey

dawiddworak88 (approved):

PR #99, opened 2022-06-20
Fixed media organisation seeds

SebaKas21:


be/src/Project/Services/Media/Media.Api/Definitions/MediaConstants.cs
```diff
@@ -7,6 +7,7 @@ public static class MediaConstants
         public struct General
         {
             public static readonly string ContainerName = "giuru";
+            public static readonly Guid DefaultOrganisationId = Guid.Parse("09affcc9-1665-45d6-919f-3d2026106ba1");
```
dawiddworak88: **suggestion:** I'd rather move and read it from the environment variable (docker-compose)

dawiddworak88 (changes requested):

dawiddworak88 (approved):

PR #108, opened 2022-07-06
Invalid password notification

SebaKas21:


fe/src/project/Account/areas/Accounts/components/SignIn/SignInForm.js
```diff
@@ -102,7 +105,9 @@ SignInForm.propTypes = {
     registerLabel: PropTypes.string.isRequired,
     developersEmail: PropTypes.string.isRequired,
     contactText: PropTypes.string.isRequired,
-    registerUrl: PropTypes.string.isRequired
+    registerUrl: PropTypes.string.isRequired,
+    isInvalidPassword: PropTypes.bool,
```
dawiddworak88: **suggestion:** There should be only one prop on the FE "ErrorMessage". There is no need to create a separate flag isInvalidPassword for it and make it so specific. Just display the error message if the ErrorMessage prop is set.

dawiddworak88 (changes requested):

be/src/Project/Services/Identity/Identity.Api/Areas/Accounts/Resources/AccountResources.en.resx
```diff
@@ -147,6 +147,9 @@
   <data name="ForgotPasswordLabel" xml:space="preserve">
     <value>Password reset</value>
   </data>
+  <data name="InvalidPassword" xml:space="preserve">
+    <value>Incorrect password. Please try again or click "Password reset" to reset it.</value>
```
dawiddworak88: **suggestion:** please make the error message more obscure. It should be "Incorrect e-mail or password. Please try again or click "Password reset" to reset it."

dawiddworak88 (changes requested):
We're almost there :)

dawiddworak88 (approved):

PR #117, opened 2022-08-18
Changed default files pagination

SebaKas21:
Now is 100 items per page :)


fe/src/shared/constants/FilesConstants.js
```diff
@@ -1,5 +1,5 @@
 export default class FilesConstants {
     static defaultPageSize() {
-        return 5;
+        return 100;
```
dawiddworak88: **suggestion**: could you please move this to the C# constants and inject it via props? Lets keep up with the pagination conventions we already have in many places in the platform.

dawiddworak88 (changes requested):

SebaKas21:
Hi @dawiddworak88 

I propose to close this PR because I am putting it in #118

PR #118, opened 2022-08-19
Added search bar for download center files component

SebaKas21:


fe/src/project/AspNetCore/shared/components/DownloadCenterFiles/DownloadCenterFiles.js
```diff
@@ -86,19 +101,47 @@ const DownloadCenterFiles = (props) => {
     const handleChangePage = (event, newPage) => {
         const startDisplayFiles = newPage * FilesConstants.defaultPageSize();
         setPage(newPage)
-        setFiles(props.files.slice(startDisplayFiles, startDisplayFiles + FilesConstants.defaultPageSize()))
+
+        let filesToSearch = props.files
```
dawiddworak88: **suggestion:** such assignments make no sense when it's used only once. Please use props.files directly.

fe/src/project/AspNetCore/shared/components/DownloadCenterFiles/DownloadCenterFiles.js
```diff
@@ -86,19 +101,47 @@ const DownloadCenterFiles = (props) => {
...
+
+        setFiles(filesToSearch.slice(startDisplayFiles, startDisplayFiles + FilesConstants.defaultPageSize()))
+    }
+
+    const handleOnChange = (event) => {
+        setSearchTerm(event.target.value);
+    }
+
+    const handleSearchButton = () => {
+        const filesFromSearch = props.files.filter(x => x.name.toLowerCase().startsWith(searchTerm.toLowerCase()));
```
dawiddworak88: **suggestion:** we need to perform search in the BE and return results to the FE. It won't work with the pagination because you will have only a subset of all files here.

dawiddworak88 (changes requested):

fe/src/project/AspNetCore/shared/components/DownloadCenterFiles/DownloadCenterFiles.js
SebaKas21: Hi @dawiddworak88 

Pagination is working, but you prefer be?

If true, i must add endpoints everywhere for files

dawiddworak88: Yes, let's do it :)

SebaKas21:
@dawiddworak88 

In this PR, I am correcting the pagination for the download center, for the rest I will create a new PR, so that it will not be wrong :)

dawiddworak88 (approved):

PR #121, opened 2022-09-09
Replaced response codes

SebaKas21:


be/src/Infrastructure/Localization/Foundation.Localization/Resources/ClientResources.de.resx
```diff
@@ -138,7 +138,7 @@
   <data name="ClientAlreadyExists" xml:space="preserve">
     <value>Der Kunder mit dieser E-Mail-Adresse wurde schon erstellt.</value>
   </data>
-  <data name="ClientApplicationNotFound" xml:space="preserve">
+  <data name="ClientApplicationNoContent" xml:space="preserve">
```
dawiddworak88: Hi @SebaKas21 , we don't want to replace the resource keys because CientApplicationNoContent makes no much sense.

dawiddworak88 (changes requested):
Hi @SebaKas21 , could you update the documentation by the way?

be/src/Infrastructure/Localization/Foundation.Localization/Resources/ClientResources.de.resx
SebaKas21: Hi @dawiddworak88,

We want only change codes (HttpStatusCode.NoContent) and messages without replace the resource keys?

be/src/Project/Web/Seller/Seller.Web/Areas/Products/Repositories/CategoriesRepository.cs
```diff
@@ -222,7 +222,7 @@ public async Task<CategorySchema> GetCategorySchemaAsync(string token, string la
                 return response.Data;
             }
 
-            if (!response.IsSuccessStatusCode && response.StatusCode == HttpStatusCode.NotFound)
+            if (!response.IsSuccessStatusCode && response.StatusCode == HttpStatusCode.NoContent)
```
dawiddworak88: **suggestion:** Hi @SebaKas21 , this requires adjustment because the 204 is considered as a success code in contrast to the not found. Please check other places if there isn't anything like that. It should be simply:

if (response.StatusCode == HttpStatusCode.NoContent)

dawiddworak88 (changes requested):

dawiddworak88 (approved):
👍

PR #122, opened 2022-09-09
Added QRCode dialog for catalog component

SebaKas21:
Hi @dawiddworak88 

If you have any comments then:
I had to add a link in the download center with and without cdn, so as not to use it: D

There is null, but without it we won't get a thumbnail, anyway take a look :)


SebaKas21:
Hi

I couldn't wait to add these fixes because I wanted everything to work in production  :D

dawiddworak88 (changes requested):
Hi @SebaKas21 ,

I've rebuilt FE but I couldn't see the qr button anywhere. Could you please have a look at it?

image

Cheers,
Dawid

SebaKas21:
Hi @dawiddworak88

 I have, without re-build

obraz

dawiddworak88 (approved):

PR #134, opened 2022-09-28
Fixed problem with resources

SebaKas21:


SebaKas21:
Hi @dawiddworak88 

I don't think that's a problem, probably it was also caused by VS update

dawiddworak88 (approved):

PR #136, opened 2022-09-30
Fixed docker-compose launch settings that started to break building solution

dawiddworak88:
It probably started to cause problems after one of the Docker or Visual Studio updates.


SebaKas21:
Hi @dawiddworak88 

I think this solution should work, we will see

SebaKas21 (approved):

PR #137, opened 2022-10-03
Fixed the errors with building Visual Studio

dawiddworak88:
Hi @SebaKas21 , Hi @dszynk ,

it should solve the problem once for good. I had to change the build order of docker-compose, so that all the API projects are the dependencies of it. Thanks to it docker-compose is built at the end and all the dlls are in the right directories. 

In the future, when new API or Web projects are created they should be checked as well:

image

image

I consolidated the nuget references and updated the Dockerfiles by the way.

Cheers,
Dawid


SebaKas21:
Hi @dawiddworak88 

Work perfectly 🚀 
You're god :D

SebaKas21 (approved):

PR #138, opened 2022-10-04
AuthenticationHelper for OrderForm

SebaKas21:
Hi @dawiddworak88 

Generally, Buyer and Seller use Foundation.Account so they are no different here. The only difference is the cookie policy but I don't think that's a problem.


dawiddworak88 (approved):
👍

PR #139, opened 2022-10-04
Unifying the nomenclature

SebaKas21:


SebaKas21:
Hi @dawiddworak88 

I analyzed the structure and we already have OrderPage, which is an order form for Seller.Web and Buyer.Web.
We used to have StatusPage and now OrderStatusPage so it's probably not a drastic change

PR #158, opened 2022-10-22
Displaying missing status on Buyer.Web

SebaKas21:


dawiddworak88 (approved):
Restore energy for upcoming Monday :) Cheers 🥇

PR #168, opened 2022-10-28
Fixed CarouselGrid styles on product detail page

SebaKas21:


fe/src/project/AspNetCore/areas/Products/components/ProductDetail/ProductDetail.js
```diff
@@ -215,7 +215,7 @@ function ProductDetail(props) {
                     labels={props.sidebar}
                 />
             </div>
-            <CarouselGrid items={props.productVariants}/>
+            <CarouselGrid items={props.productVariants} className="pt-6"/>
```
dawiddworak88, line 218: **suggestion:** there is a padding-top: 0; in carousel grid css.  Maybe it would be enough to delete this line? Would it break something?

dawiddworak88 (changes requested):

fe/src/project/AspNetCore/areas/Products/components/ProductDetail/ProductDetail.js
SebaKas21: This will revert to our earlier change to the spacing between sections on the home page

dawiddworak88 (approved):
👍

PR #172, opened 2022-10-28
Fixed buttons order for client form

SebaKas21:
Before:

obraz


SebaKas21:
Now (for example): 

obraz

dawiddworak88 (approved):
👍

PR #177, opened 2022-11-03
Meaningful error message for expired reset password link

SebaKas21:


SebaKas21:
Hi @dawiddworak88 This problem was also fixed in recently merged PR. I close and clear our list of opened pr :D

PR #178, opened 2022-11-04
Extended z-indexes on other components

SebaKas21:


dawiddworak88:
@SebaKas21 , as discussed we need less error prone solution.

PR #181, opened 2022-11-07
Added missing countries to list of countries

SebaKas21:


SebaKas21:
Hi @dawiddworak88 

I am thinking to redo the list of countries, because the ID is missing and there will be a problem with the translation: D

dawiddworak88 (approved):

PR #183, opened 2022-11-09
Fixed canceled order item statuses

SebaKas21:


dawiddworak88 (approved):
Looks good 👍

be/src/Project/Services/Ordering/Ordering.Api/Services/OrdersService.cs
```diff
@@ -864,20 +864,23 @@ private async Task MapStatusesToOrderStatusId(Guid? orderId)
...
-                            OrderItemStatusId = lastOrderItemStatusChange.OrderItemStatusId
+                            OrderItemStatusId = lastOrderItemStatusChange.OrderItemStatusId,
+                            CreatedDate = lastOrderItemStatusChange.CreatedDate
                         });
                     }
                 }
 
                 bool isSameStatus = lastOrderItemStatusChanges.DistinctBy(x => x.OrderItemStatusId).Count() == 1;
 
+                var lastOrderItemStatus = lastOrderItemStatusChanges.FirstOrDefault();
```
dawiddworak88, line 874: **suggestion:** let's add a null check for lastOrderItemStatus . If it's null there will be an exception on line 879-880.

dawiddworak88 (changes requested):

be/src/Project/Services/Ordering/Ordering.Api/Services/OrdersService.cs
```diff
@@ -864,20 +864,23 @@ private async Task MapStatusesToOrderStatusId(Guid? orderId)
                         lastOrderItemStatusChanges.Add(new OrderItemStatusServiceModel
                         {
                             OrderItemStateId = lastOrderItemStatusChange.OrderItemStateId,
-                            OrderItemStatusId = lastOrderItemStatusChange.OrderItemStatusId
+                            OrderItemStatusId = lastOrderItemStatusChange.OrderItemStatusId,
+                            CreatedDate = lastOrderItemStatusChange.CreatedDate
```
dawiddworak88: **question:** wouldn't CreatedDate = DateTime.UtcNow be enough?

dawiddworak88 (changes requested):

be/src/Project/Services/Ordering/Ordering.Api/Services/OrdersService.cs
SebaKas21: it was for testing, i removed

dawiddworak88 (approved):
👍

PR #191, opened 2022-11-23
Improvements for importing order items from excel on Buyer.Web

SebaKas21:


dawiddworak88:
Hi @SebaKas21 , I'm closing the issue because I could fix it with the following change. The key to fix it was to reproduce the issue by removing the basket cookie in the browser before uploading a file and placing an order:
https://github.com/dawiddworak88/Giuru/commit/615f60afdae0301f71512350157e5a6a6cceaaf3

PR #193, opened 2022-11-29
Fixed missing products on inventory

SebaKas21:


dawiddworak88 (changes requested):
@SebaKas21, could you make a hotfix out of main and create a PR there? Would it be possible to make this search case insensitive (toLowercase())?

SebaKas21:
Hi @dawiddworak88 

Okay, in this case, it searches well regardless of case

PR #195, opened 2022-11-29
Fixed sku case insensitive for odering file

SebaKas21:


dawiddworak88:
Hi @SebaKas21 , match would be very dangerous because it performs fuzzy search. Fuzzy search means it could give 'similar' results which would return wrong results with skus. Searching by SKU must be an exact search. You'll find the details here: https://stackoverflow.com/questions/23150670/elasticsearch-match-vs-term-query

PR #197, opened 2022-12-02
Fixes related to pagination constraints

SebaKas21:
Hi @dawiddworak88 

I think I need to remove unnecessary validators in controllers, what do you think?


dawiddworak88:
Yes :)

W dniu pt., 2.12.2022 o 13:37 Sebastian Kasperek ***@***.***>
napisał(a):

be/src/Infrastructure/GenericRepository/Foundation.GenericRepository/Definitions/Constants.cs
```diff
@@ -5,5 +5,7 @@ public static class Constants
         public static readonly int DefaultPageIndex = 1;
         public static readonly int DefaultItemsPerPage = 25;
         public static readonly int MaxItemsPerPage = 100;
+        public static readonly int MaxItemsPerPageLimit = 1000;
```
dawiddworak88: **suggestion:** let's change it to 2000

be/src/Infrastructure/GenericRepository/Foundation.GenericRepository/Definitions/Constants.cs
```diff
@@ -5,5 +5,7 @@ public static class Constants
...
+        public static readonly int EmptyTotal = 0;
```
dawiddworak88: **suggestion:** let's remove it completely

be/src/Project/Services/Catalog/Catalog.Api/Services/Categories/CategoriesService.cs
```diff
@@ -51,7 +52,18 @@ public async Task<PagedResults<IEnumerable<CategoryServiceModel>>> GetAsync(GetC
 
             categories = categories.ApplySort(model.OrderBy);
 
-            var pagedResults = categories.PagedIndex(new Pagination(categories.Count(), model.ItemsPerPage), model.PageIndex);
+            var pagedResults = categories.PagedIndex(new Pagination(Constants.EmptyTotal, Constants.MaxItemsPerPageLimit), Constants.DefaultPageIndex);
```
dawiddworak88: **suggestion:** please replace with the following everywhere:

`PagedResults<IEnumerable<Category>> pagedResults;`

dawiddworak88 (changes requested):

be/src/Project/Services/Catalog/Catalog.Api/Services/Categories/CategoriesService.cs
SebaKas21: Hi @dawiddworak88 as variable?

be/src/Project/Services/Client/Client.Api/v1/Controllers/ClientsApplicationsController.cs
```diff
@@ -290,41 +290,35 @@ public async Task<IActionResult> Get(string ids, string searchTerm, int pageInde
...
+                        })
+                    };
 
-                        return this.StatusCode((int)HttpStatusCode.OK, response);
-                    }
+                    return this.StatusCode((int)HttpStatusCode.OK, response);
                 }
 
-                throw new CustomException(string.Join(ErrorConstants.ErrorMessagesSeparator, validationResult.Errors.Select(x => x.ErrorMessage)), (int)HttpStatusCode.UnprocessableEntity);
+                throw new CustomException("", (int)HttpStatusCode.UnprocessableEntity);
```
dawiddworak88: **suggestion:** let's always simply return  this.StatusCode((int)HttpStatusCode.UnprocessableEntity); if there is no error message.

dawiddworak88 (changes requested):
We're almost there ⭐

dawiddworak88 (approved):
Thank you 🥇

PR #204, opened 2022-12-15
Optimized getting categories in Catalog.Api

dawiddworak88:
In this PR I reduced the number of database calls to get categories to three. It takes 7 ms in debugging mode to get them on local environment. The goal is to implement similar mechanism in all services (products, orders). It will significantly improve the tracing metrics and in effect the page loading times, pressure on the database and user experience. 

I've also started using _ for private properties since it's a Microsoft guideline and removed 'this.' because it's redundant.

### DIFF
--- a/be/src/Project/Services/Catalog/Catalog.Api/v1/Categories/Controllers/CategoriesController.cs
+++ b/be/src/Project/Services/Catalog/Catalog.Api/v1/Categories/Controllers/CategoriesController.cs
@@ -65,7 +65,7 @@ namespace Catalog.Api.v1.Categories.Controllers
                 OrderBy = orderBy
             };
 
-            var categories = _categoryService.GetAsync(serviceModel);
+            var categories = _categoryService.Get(serviceModel);
 
             if (categories is not null)
             {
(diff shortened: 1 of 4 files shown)


SebaKas21:
Hi @dawiddworak88 

I understand, we use the full sql potential and download everything for fix it

be/src/Project/Services/Catalog/Catalog.Api/v1/Categories/Controllers/CategoriesController.cs
```diff
@@ -51,7 +52,7 @@ public CategoriesController(ICategoriesService categoryService)
         [ProducesResponseType((int)HttpStatusCode.OK, Type = typeof(PagedResults<IEnumerable<CategoryResponseModel>>))]
         [ProducesResponseType((int)HttpStatusCode.NoContent)]
         [ProducesResponseType((int)HttpStatusCode.UnprocessableEntity)]
-        public async Task<IActionResult> Get(string searchTerm, int? level, bool? leafOnly, int? pageIndex, int? itemsPerPage, string orderBy)
```
SebaKas21, line 54: **question:** Why we stop using async?

dawiddworak88: Hi @SebaKas21, it makes no sense to use async if inside the method there is no await inside. It throws a warning in Visual Studio, makes building solution slower and execution of the code less performant.

dawiddworak88:
Hi @SebaKas21, could you please review my changes? We'll need to apply similar ones to News, Products, Orders, etc.

PR #219, opened 2023-01-11
Optimization for services

SebaKas21:


be/src/Project/Services/DownloadCenter/DownloadCenter.Api/v1/Controllers/CategoriesController.cs
```diff
@@ -159,11 +159,11 @@ public async Task<IActionResult> Get(Guid? id)
             };
 
             var validator = new GetCategoryModelValidator();
-            var validationResult = await validator.ValidateAsync(serviceModel);
```
dawiddworak88: **suggestion:** we prefer async methods. If they are available or used, let's leave them

be/src/Project/Services/DownloadCenter/DownloadCenter.Api/Services/DownloadCenter/DownloadCenterService.cs
```diff
@@ -223,28 +223,24 @@ public async Task<DownloadCenterCategoryServiceModel> GetDownloadCenterCategoryA
...
 
-            var translations = _context.DownloadCenterCategoryTranslations.Where(x => x.CategoryId == downloadCenterCategory.Id || x.CategoryId == downloadCenterCategory.ParentCategoryId || subcategories.Select(y => y.Id).Contains(x.CategoryId)).ToList();
-
-            var files = _context.DownloadCenterCategoryFiles.Where(x => x.CategoryId == downloadCenterCategory.Id && x.IsActive);
-
             return new DownloadCenterCategoryServiceModel
             {
                 Id = downloadCenterCategory.Id,
-                CategoryName = translations.FirstOrDefault(t => t.CategoryId == model.Id && t.IsActive && t.Language == model.Language)?.Name ?? translations.FirstOrDefault(t => t.CategoryId == model.Id && t.IsActive)?.Name,
+                CategoryName = downloadCenterCategory.Translations.FirstOrDefault(t => t.CategoryId == model.Id && t.IsActive && t.Language == model.Language)?.Name ?? downloadCenterCategory.Translations.FirstOrDefault(t => t.CategoryId == model.Id && t.IsActive)?.Name,
```
dawiddworak88, line 231: **suggestion:** if you don't add .Include and .AsSingleQuery statements it will result in many database calls

dawiddworak88 (changes requested):

dawiddworak88 (approved):

PR #225, opened 2023-01-23
Adding new fields for product cards

SebaKas21:
HI @dawiddworak88 

Adding new fields at the moment is primary, but in next step i want add modal with all options and definitions.

### DIFF
--- a/fe/src/project/Seller.Portal/areas/Products/components/ProductCardForm/ProductCardForm.js
+++ b/fe/src/project/Seller.Portal/areas/Products/components/ProductCardForm/ProductCardForm.js
@@ -189,6 +189,48 @@ const ProductCardForm = (props) => {
         updateSchema(newElements, schema)
     }
 
+    const getIdFromElements = (elements) => {
+        const names = elements.map((element) => element.name);
+        const defaultNameLength = props.defaultInputName.length;
+      
+        return names.length > 0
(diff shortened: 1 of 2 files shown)


fe/src/project/Seller.Portal/areas/Products/components/ProductCardForm/ProductCardForm.js
```diff
@@ -189,6 +189,48 @@ const ProductCardForm = (props) => {
         updateSchema(newElements, schema)
     }
 
+    const getIdFromElements = (elements) => {
```
dawiddworak88, line 192: **suggestion:** wouldn't it better to generate a random id? https://stackoverflow.com/questions/105034/how-do-i-create-a-guid-uuid

dawiddworak88 (changes requested):

fe/src/project/Seller.Portal/areas/Products/components/ProductCardForm/ProductCardForm.js
SebaKas21: Hey, this is so that you can create more new fields, e.g. NewElement1, NewElement2 etc.

I want to use guid for definitions and their values

dawiddworak88 (approved):

PR #230, opened 2023-01-30
Fixed endpoint for getting product by sku

SebaKas21:


dawiddworak88 (changes requested):
Hi Sebastian, it should actually work with the keyword. Could you check whether it finds product with keyword suffix when it's lowercase, etc.?

SebaKas21:
Hi @dawiddworak88 

I checked in the code and by writing the query manually. Success was true, but no data,

Without keyword searches in both cases

SebaKas21:
@dawiddworak88 

What you think?

dawiddworak88:
I fixed it by removing to lowercaseinvariant. Once we update the Elasticsearch version we'll add support for case insensitivity

PR #233, opened 2023-01-30
Extended Analytics.Api for Seller.Web

SebaKas21:


be/src/Project/Services/Analytics/Analytics.Api/Services/SalesAnalytics/SalesService.cs
```diff
@@ -52,31 +52,47 @@ public async Task CreateAsync(CreateSalesAnalyticsServiceModel model)
...
                     await _context.ClientDimensions.AddAsync(clientDimension.FillCommonProperties());
+                } 
+                else
+                {
+                    clientDimension.Name = client.Name;
+                    clientDimension.OrganisationId = client.OrganisationId;
                 }
 
-                var locationDimension = await _context.LocationDimensions.FirstOrDefaultAsync(x => x.CountryId == client.CountryId);
+                Nullable<Guid> locationDimensionId = null;
```
dawiddworak88: **suggestion:** Guid?

dawiddworak88 (commented):
Hi @SebaKas21 , could you check how much time it takes to open the page using C# StopWatch and Network tab in Chrome browser?

SebaKas21:
Hi @dawiddworak88 

Can you describe on why check page uptime if it's just an API extension?

be/src/Project/Services/Analytics/Analytics.Api/Services/SalesAnalytics/SalesService.cs
SebaKas21: I will think about it

dawiddworak88 (approved):

PR #245, opened 2023-02-10
Media-Api build and deployment files

SebaKas21:


misc/deployment/variables.yml
```diff
@@ -5,3 +5,5 @@ variables:
   value: 'b2b-global-api'
 - name: b2bNewsApiImageRepository
```
dawiddworak88: **suggestion:** please move service variables to variables.yml files in service directories. We don't want to have references to services in global variables.yml.

dawiddworak88 (approved):
👍

PR #261, opened 2023-02-23
Fixed displaying countries in chart

SebaKas21:
Hi @dawiddworak88 

I noticed that countries were not showing up 3 months back

### DIFF
--- a/be/src/Project/Services/Analytics/Analytics.Api/Services/SalesAnalytics/SalesService.cs
+++ b/be/src/Project/Services/Analytics/Analytics.Api/Services/SalesAnalytics/SalesService.cs
@@ -219,9 +219,14 @@ namespace Analytics.Api.Services.SalesAnalytics
 
         public IEnumerable<CountrySalesServiceModel> GetCountrySales(GetCountriesSalesServiceModel model)
         {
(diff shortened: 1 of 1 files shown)


be/src/Project/Services/Analytics/Analytics.Api/Services/SalesAnalytics/SalesService.cs
```diff
@@ -219,9 +219,14 @@ select new
 
         public IEnumerable<CountrySalesServiceModel> GetCountrySales(GetCountriesSalesServiceModel model)
         {
+            var endDate = DateTime.UtcNow;
+            var startDate = endDate.AddMonths(-2);
```
dawiddworak88, line 223: **suggestion:** exactly this we need to avoid. We should inject it dynamically via parameters

dawiddworak88 (approved):

be/src/Project/Services/Analytics/Analytics.Api/Services/SalesAnalytics/SalesService.cs
SebaKas21: I know and remember.  This will be in the charts-data-range branch

PR #269, opened 2023-03-15
Website pipelines 

SebaKas21:
Hi @dawiddworak88 

Can you look on build pipelines for npm tasks?


misc/deployment/BuyerWeb/helm-chart/templates/deployment.yaml
```diff
@@ -0,0 +1,83 @@
+apiVersion: apps/v1
```
dawiddworak88, line 1: **suggestion:** In the deployment.yaml files there are missing 'buyer-web-react-ssr' missing. Upon deployments both of the services the buyer-web and buyer-web-react be deployed together because it's one app

misc/deployment/BuyerWeb/azure-deployment-build-pipeline.yml
```diff
@@ -0,0 +1,55 @@
+trigger:
```
dawiddworak88, line 1: **suggestion:** it should build and push both buyer-web and buyer-web-react-ssr  apps

dawiddworak88 (changes requested):

misc/deployment/BuyerWeb/azure-deployment-build-pipeline.yml
SebaKas21: Okay, I will try today

SebaKas21: Should I add it in the same task or another?

dawiddworak88 (approved):

PR #280, opened 2023-06-25
Redesigned footer

SebaKas21:
Hi @dawiddworak88,

Footer has been redesigned according to the design in figma

### DIFF
--- a/be/src/Infrastructure/Localization/Foundation.Localization/Resources/GlobalResources.de.resx
+++ b/be/src/Infrastructure/Localization/Foundation.Localization/Resources/GlobalResources.de.resx
@@ -166,7 +166,7 @@
     <value>Link kopieren</value>
   </data>
   <data name="Copyright" xml:space="preserve">
-    <value>Copyright © [year]</value>
+    <value>© [year] | ELTAP Spółka z ograniczoną odpowiedzialnością Sp.k. Alle Rechte vorbehalten.</value>
   </data>
   <data name="CreatedDate" xml:space="preserve">
     <value>Erstellungsdatum</value>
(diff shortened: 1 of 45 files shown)


be/src/Infrastructure/Localization/Foundation.Localization/Resources/GlobalResources.de.resx
```diff
@@ -166,7 +166,7 @@
     <value>Link kopieren</value>
   </data>
   <data name="Copyright" xml:space="preserve">
-    <value>Copyright © [year]</value>
+    <value>© [year] | ELTAP Spółka z ograniczoną odpowiedzialnością Sp.k. Alle Rechte vorbehalten.</value>
```
dawiddworak88, line 169: **suggestion:** lets' move these texts to Content.API because when I deploy it to other tenant the text is going to be invalid

dawiddworak88 (changes requested):

SebaKas21:
Everything has been corrected

be/src/Infrastructure/Localization/Foundation.Localization/Resources/GlobalResources.de.resx
dawiddworak88, line 169: **suggestion:** Hi @SebaKas21, there is 'ELTAP' hardcoded. The place of this text is either in the content API. In my view the easiest will be to revert the previous version

dawiddworak88 (changes requested):

dawiddworak88:
@SebaKas21, here it's also important to add the required content API postgres SQL scripts to the init script. Upon first checkout and project run the footer should be present and look nice

dawiddworak88 (approved):
Thank you

PR #294, opened 2023-10-20
Client delivery adresses on BE

SebaKas21:
Hi @dawiddworak88 

In this PR is only functionality for BE


PiotrZaranek (approved):

be/src/Project/Services/Client/Client.Api/Infrastructure/Clients/Entities/Address.cs
```diff
@@ -6,12 +6,6 @@ namespace Client.Api.Infrastructure.Clients.Entities
 {
     public class Address : Entity
     {
-        public string Company { get; set; }
```
dawiddworak88: **suggestion:** let's keep these fields because in the address can contain a name of the company or recipient first name / last name

be/src/Project/Services/Client/Client.Api/Infrastructure/Clients/Entities/Client.cs
```diff
@@ -27,5 +27,7 @@ public class Client : Entity
 
         [Required]
         public Guid SellerId { get; set; }
+
+        public Guid? DefaultAddressId { get; set; }
```
dawiddworak88: **suggestion:** please rename to DefaultDeliveryAddressId

dawiddworak88 (changes requested):

be/src/Project/Services/Client/Client.Api/Infrastructure/Clients/Entities/Address.cs
SebaKas21: Okay, but what you think to set the property name as Recipient?

dawiddworak88 (approved):
Thank you 🌞

PR #317, opened 2024-01-08
Fix for categories endpoint

SebaKas21:


dawiddworak88:
Hi @SebaKas21, it would make the situation worse because it loads categories doesn't matter what language. It shouldn't make any difference to call ToList(). Probably because you're running it locally, it works faster. I'll have a look into it.

PR #319, opened 2024-01-09
Reverted schemas duplication

SebaKas21:
Hi @dawiddworak88 ,

In my opinion, it was enough to just remove the display of schemas from the /Categories endpoint and leave the rest. So I did that and restored it because the schema duplication uses completely different endings

The removal of schema display itself is noticeable in the local environment, so it should probably be okay


dawiddworak88 (approved):
OK, thank you. Lets test

PR #324, opened 2024-01-23
Deactivation and activation of customer accounts

SebaKas21:


dawiddworak88 (changes requested):
Hi @SebaKas21, it looks good in general. Let's change the column name to IsDisabled instead of IsActive and adjust the logic according to it. By default the accounts will be enabled and working thanks to it

SebaKas21:
Hi @dawiddworak88 

You mean IsDisabled instead of IsActive in identity api? or also in client api?

dawiddworak88 (approved):
Thank you @SebaKas21 🥇

PR #335, opened 2024-03-18
Fixed saving clients with dynamic fields

SebaKas21:


dawiddworak88 (approved):
🤝 Thank you @SebaKas21

PR #345, opened 2024-04-25
Fixed gap in access to information about the organization

SebaKas21:


be/src/Project/Services/Identity/Identity.Api/v1/Controllers/OrganisationsController.cs
```diff
@@ -39,7 +39,6 @@ public OrganisationsController(IOrganisationService organisationService)
         /// <returns>The seller.</returns>
         [HttpGet, MapToApiVersion("1.0")]
         [Route("{email}")]
-        [AllowAnonymous]
```
dawiddworak88, line 42: Are you sure? What is the difference?

SebaKas21: I would like to force authorization, because without it someone would be able to check whether a given email exists and generally get information about a given organization

dawiddworak88:
Hi @SebaKas21, let's deploy it to the testing env and check whether it doesn't break basic functionalitites

SebaKas21:
Hi @dawiddworak88 this solution is only on seller.web for creating new clients. All my tests passed

dawiddworak88:
Let's leave it for now. I'm concerned the existing API integrations would break

PR #346, opened 2024-04-26
Fixed localizedAreaSlugRoute order in buyer.web

SebaKas21:


dawiddworak88 (commented):
I'd suggest prefixing slug with /slug/. The current implementation is risky because imagine any of the areas like Content has the same name as Slug

PiotrZaranek (approved):

PR #360, opened 2024-06-12
Improved reading of acr name from secrets

SebaKas21:
Hi everyone,

I decided to add this fix because some acr names are read directly from the file and some from secrets, so I decided to unify


dawiddworak88 (approved):
🥇 thank you

MarcinKalinowski1 (approved):
approved

PR #363, opened 2024-06-19
Changed DEV to TEST and updated path to gitops overlays

SebaKas21:


MarcinKalinowski1 (approved):

dawiddworak88 (approved):
You can @SebaKas21 change names of the files too, so that they contain test instead of dev :)

PR #377, opened 2024-09-02
added SKU to parent product name to avoid confusion (multiple product…

MarcinKalinowski1:
hey, there may be more than one product with the same name in the "Select parent product" field. I suggest adding a SKU next to the name so that the user can distinguish it and assign it correctly.


dawiddworak88 (changes requested):
Hi @MarcinKalinowski1, let's have the endpoint also updated so that SKU is visible after obtaining search results as well

dawiddworak88 (approved):
🎖️

PR #385, opened 2024-09-24
Add a check to verify if the attribute is used before deleting.

PiotrZaranek:


be/src/Project/Web/Seller/Seller.Web/Areas/Products/Repositories/CategoriesRepository.cs
```diff
@@ -250,5 +251,30 @@ public async Task SaveAsync(string token, string language, Guid? id, CategorySch
...
         }
+
+        public async Task<bool> CategoriesSchemasImplementAttibuteAsync(string token, string language, Guid? attributeId)
+        {
+            var apiRequest = new ApiRequest<RequestModelBase>
+            {
+                Language = language,
+                Data = new RequestModelBase(),
+                AccessToken = token,
+                EndpointAddress = $"{_settings.Value.CatalogUrl}{ApiConstants.Catalog.CategorySchemasApiEndpoint}/ImplementAttribute/{attributeId}"
```
SebaKas21: **suggestion:** Add new variable to ApiConstants with value: /api/v1/CategorySchemas/ImplementAttribute and meaningful name

SebaKas21 (approved):
Hi @dawiddworak88 @PiotrZaranek 

I've added a minor fix regarding the repository. In terms of functionality it works as intended

be/src/Project/Services/Catalog/Catalog.Api/v1/Categories/Controllers/CategoriesController.cs
```diff
@@ -407,5 +407,43 @@ public async Task<IActionResult> GetCategorySchemasByCategoryId(Guid? categoryId
...
             throw new CustomException(string.Join(ErrorConstants.ErrorMessagesSeparator, validationResult.Errors.Select(x => x.ErrorMessage)), (int)HttpStatusCode.UnprocessableEntity);
         }
+
+        /// <summary>
+        /// Gets information on whether the attribute is in use.
+        /// </summary>
+        /// <param name="id">The attribute id.</param>
+        /// <returns>The result of the check. (true or false)</returns>
+        [HttpGet, MapToApiVersion("1.0")]
+        [Route("CategorySchemas/ImplementAttribute/{id}")]
```
dawiddworak88: **suggestion:** Adjust the route to be CategorySchemas/AttributeInUse/{key}

be/src/Project/Services/Catalog/Catalog.Api/Services/Categories/CategoriesService.cs
```diff
@@ -379,6 +379,20 @@ public async Task<CategorySchemasServiceModel> GetCategorySchemasAsync(GetCatego
             };
         }
 
+        public async Task<bool> CategoriesSchemasImplementAttributeAsync(CategoriesSchemasImplementAttributeServiceModel model)
+        {
+            var query = $@"
```
dawiddworak88, line 384: **suggestion:** can't we use the almost exact query that I sent via email? In my view it won't work for example when one deletes the 'length' key because it doesn't have any definitions, it's just a number

dawiddworak88 (changes requested):

PR #402, opened 2024-10-18
Reduced exceptions in availableproducts endpoint

SebaKas21:


dawiddworak88 (approved):
Please create it as hotfix to main too @SebaKas21

PR #423, opened 2025-02-19
Correct styles for markdown

PiotrZaranek:


SebaKas21 (approved):
Nice 👍 

@PiotrZaranek please just check twice times if everywhere markdown.scss file is imported where is required

PR #426, opened 2025-03-04
Add column "in total" in new order form 

PiotrZaranek:


dawiddworak88 (approved):
It would be great to have the same on the seller side (in the next PR)

fe/projects/AspNetCore/src/areas/Orders/components/NewOrderForm/NewOrderForm.js
```diff
@@ -514,6 +515,7 @@ function NewOrderForm(props) {
                                                         <TableCell>{item.quantity}</TableCell>
                                                         <TableCell>{item.stockQuantity}</TableCell>
                                                         <TableCell>{item.outletQuantity}</TableCell>
+                                                        <TableCell className="has-text-weight-bold">{item.quantity + item.stockQuantity + item.outletQuantity}</TableCell>
```
SebaKas21: **suggestion:** create additional function getTotalQuantities instead of adding each one separately

SebaKas21 (approved):

PR #439, opened 2025-05-14
Added prices to basket and ordering api

SebaKas21:


be/src/Project/Web/Buyer/Buyer.Web/Shared/Services/Prices/PriceService.cs
```diff
@@ -55,20 +54,27 @@ public async Task<Price> GetPrice(
...
+            try
             {
-                if (response.Data.Amount is null)
-                {
-                    return null;
-                }
+                var response = await _apiClientService.PostAsync<ApiRequest<PriceRequestModel>, PriceRequestModel, PriceResponseModel>(apiRequest);
 
-                return new Price
+                if (response.IsSuccessStatusCode && response.Data != null)
```
PiotrZaranek, line 62: **Suggestion** (non-blocking): Please use `is not` intead of `!=`

PiotrZaranek (approved):
Well done 👍

PR #452, opened 2025-06-14
Added max allowed order quantity

SebaKas21:


PiotrZaranek (approved):

dawiddworak88 (changes requested):
Hi @SebaKas21, it would be great to check following remarks:

## ❌ __Critical Issues Found__

### 1. __Backend Validator Not Updated__

The most critical issue is that the current `UpdateBasketModelValidator.cs` file doesn't contain the changes shown in the diff. The validator still uses the old constructor without the `maxAllowedOrderQuantity` parameter and lacks the quantity limit validation.

__Current code:__

```csharp
public UpdateBasketModelValidator()
```

__Expected code (from diff):__

```csharp
public UpdateBasketModelValidator(int? maxAllowedOrderQuantity)
```

### 2. __Controller Not Updated__

The `BasketsController.cs` still instantiates the validator without passing the configuration:

```csharp
var validator = new UpdateBasketModelValidator(); // Missing parameter
```

Should be:

```csharp
var validator = new UpdateBasketModelValidator(_options.Value.MaxAllowedOrderQuantity);
```

### 3. __Missing Dependency Injection__

The controller doesn't inject `IOptions<AppSettings>` as shown in the diff.

## ⚠️ __Potential Issues__

### 1. __Inconsistent Error Messages__

- Backend validation uses a generic message with the limit value
- Frontend uses a localized message without showing the actual limit
- Consider showing the limit (1000) in the frontend error message for better UX

### 2. __Configuration Validation__

- No validation that `MaxAllowedOrderQuantity` is a positive number
- Should add validation in startup/configuration

### 3. __Frontend Validation Logic__

The frontend validation correctly calculates total quantity:

```javascript
const totalQuantity = parseInt(quantity) + parseInt(stockQuantity) + parseInt(outletQuantity);
```

However, there's potential for parseInt() to return NaN if values are empty strings.

### 4. __Missing PropTypes__

Frontend components are missing PropTypes for the new properties:

- `maxAllowedOrderQuantity`
- `maxAllowedOrderQuantityErrorMessage`

## 🔧 __Recommended Fixes__

### High Priority

1. __Apply the missing backend changes__ from the diff to make the validator and controller work properly
2. __Add proper dependency injection__ for IOptions in BasketsController
3. __Update validator instantiation__ to pass the configuration parameter

### Medium Priority

1. __Add configuration validation__ in Startup.cs to ensure MaxAllowedOrderQuantity is valid
2. __Improve frontend validation__ to handle NaN cases from parseInt()
3. __Add missing PropTypes__ for new frontend properties
4. __Consider showing the limit value__ in frontend error messages

### Low Priority

1. __Add unit tests__ for the new validation logic
2. __Consider making the limit configurable per client__ if business requirements change

dawiddworak88 (approved):
🎖️

PR #455, opened 2025-07-02
Add filters component and primary products sorting

PiotrZaranek:


be/src/Project/Web/Buyer/Buyer.Web/Shared/ViewModels/Catalogs/CatalogViewModel.cs
```diff
@@ -14,6 +15,7 @@ public class CatalogViewModel
         public string Title { get; set; }
         public string ResultsLabel { get; set; }
         public string NoResultsLabel { get; set; }
+        public string noFilteredResultsLabel { get; set; }
```
SebaKas21: **suggestion (non-blocking):** case sensitive

SebaKas21 (changes requested):

be/src/Infrastructure/Catalog/Foundation.Catalog/Repositories/ProductSearchRepositories/IProductSearchRepository.cs
```diff
@@ -17,8 +17,10 @@ Task<PagedResults<IEnumerable<ProductSearchModel>>> GetAsync(
             string searchTerm,
             int? pageIndex, 
             int? itemsPerPage,
-            string orderBy);
-        Task<PagedResults<IEnumerable<ProductSearchModel>>> GetAsync(string language, Guid? organisationId, IEnumerable<Guid> ids, string orderBy);
+            string orderBy,
```
dawiddworak88: suggestion: orderBy should be enough for sorting

be/src/Infrastructure/Catalog/Foundation.Catalog/Repositories/ProductSearchRepositories/IProductSearchRepository.cs
```diff
@@ -17,8 +17,10 @@ Task<PagedResults<IEnumerable<ProductSearchModel>>> GetAsync(
...
+            string filters,
```
dawiddworak88: suggestion: let's leave anything related to filters to a separate pull request

dawiddworak88 (changes requested):
I've analyzed the PR and can now provide a comprehensive pull request review focused on the sorting functionality and performance considerations. Based on my analysis, here are the key findings ranked by importance:

## **CRITICAL ISSUES (High Priority)**

### 1. **Performance Issue: Inefficient Elasticsearch Query Logic**
**Location**: `ProductSearchRepository.cs` lines 51-62
**Issue**: The sorting logic is problematic - when `sort == SortingConstants.Newest`, it adds an additional filter `IsNew = true` to the query, but this should be handled in the sorting logic, not as a query filter.

```csharp
if (sort == SortingConstants.Newest)
{
    query = query && Query<ProductSearchModel>.Term(t => t.Field(x => x.IsNew).Value(true));
}
else
{
    if (isNew.HasValue)
    {
        query = query && Query<ProductSearchModel>.Term(t => t.IsNew, isNew.Value);
    }
}
```

**Impact**: This changes the result set instead of just sorting it, which is incorrect behavior for a sorting feature.

### 2. **Inconsistent API Design**
**Location**: Multiple files - new `FilteredProductsApiEndpoint` vs existing `ProductsApiEndpoint`
**Issue**: The PR introduces a new endpoint `/api/v1/products/filtered` but the existing endpoint should handle both filtering and sorting to maintain consistency.

### 3. **Unused Parameters in Repository Methods**
**Location**: `ProductSearchRepository.cs` and `IProductSearchRepository.cs`
**Issue**: The `filters` parameter is added to method signatures but never used in the implementation, indicating incomplete functionality.

## **MAJOR ISSUES (Medium-High Priority)**

### 4. **Missing Elasticsearch Sort Optimization**
**Location**: `ProductSearchRepository.cs` `Sorting<T>` method (lines 302-315)
**Issue**: The sorting method falls back to an empty `SortDescriptor<T>()` for default sorting, which may not utilize Elasticsearch's default relevance scoring effectively.

### 5. **Frontend State Management Issues**
**Location**: `Catalog.js` 
**Issue**: The sorting and filtering state is managed locally in the component but the `filters` parameter is never actually sent to the API, making the filtering UI non-functional.

### 6. **Inconsistent Parameter Handling**
**Location**: Multiple service and repository methods
**Issue**: Some methods receive both `orderBy` and `sort` parameters, creating confusion about which takes precedence.

## **MODERATE ISSUES (Medium Priority)**

### 7. **Missing Error Handling**
**Location**: Frontend components
**Issue**: No error handling for invalid sort/filter combinations or API failures specific to the new filtering functionality.

### 8. **Localization Inconsistencies**
**Location**: Resource files
**Issue**: Some localization keys are added but not consistently used across all supported languages.

### 9. **Code Duplication**
**Location**: Multiple API controllers
**Issue**: Similar logic for handling products with sorting/filtering is duplicated across different controllers.

## **MINOR ISSUES (Low Priority)**

### 10. **Missing Documentation**
**Location**: New API endpoint
**Issue**: The new filtered endpoint lacks comprehensive API documentation for the filters parameter format.

### 11. **Inconsistent Naming**
**Location**: Various files
**Issue**: Mix of `sort` and `orderBy` parameter names across the codebase.

## **RECOMMENDATIONS**

1. **Fix the Elasticsearch Query Logic**: Remove the filtering logic from the sorting condition and handle "newest" as a pure sort operation.

2. **Implement Proper Filtering**: Complete the filtering functionality by actually using the filters parameter in the Elasticsearch queries.

3. **Optimize Elasticsearch Performance**: Ensure all sorting operations are pushed to Elasticsearch level rather than being handled in application code.

4. **Add Integration Tests**: The sorting functionality needs comprehensive testing to ensure Elasticsearch queries are optimized.

Would you like me to elaborate on any of these issues or provide specific code recommendations for the most critical problems?

dawiddworak88 (changes requested):
────────────────────────────────────────────────────────
🚨 1. Breaking API & behavioural change (HIGH)  
────────────────────────────────────────────────────────
• Buyer-side controllers now *require* a new query-string parameter `orderBy`.  
  – Existing external callers that omit the argument will encounter model-binding failure or silent change of sorting (non-deterministic ES order).  
• Default ordering has changed: previously the code path fell back to `Name ASC`; after refactor, `"default"` returns *null* and Elasticsearch falls back to `_score/_doc`, breaking paging consistency and cache keys.  
RECOMMENDATION  
  1. Keep `orderBy` optional in the public contract.  
  2. Map `"default"` (or `null`) to the previous `Name ASC` sort in `Sorting<T>` so behaviour is preserved.  
  3. Document the new parameter in Swagger and bump API version if a change is unavoidable.

────────────────────────────────────────────────────────
🚨 2. Input validation / Security (HIGH)  
────────────────────────────────────────────────────────
`orderBy` value from the client is passed straight to `ToElasticSortList<T>()`.  
• Malformed field names or malicious strings will throw or may lead to unexpected ES queries.  
RECOMMENDATION  
  – Validate against an allow-list (`SortingConstants`) before transformation; return HTTP 400 on invalid values.

────────────────────────────────────────────────────────
⚡ 3. “Newest” logic inconsistency & performance (MED-HIGH)  
────────────────────────────────────────────────────────
• For `orderBy == newest` the code adds a filter `IsNew == true` but **no sort**. Results are therefore scored, not ordered by freshness and risk returning “old” items that match search terms better.  
• Duplicate `IsNew` filter added in two query blocks.  
• Filtering reduces result set, but removing `.Sort()` makes paging unstable.  
RECOMMENDATION  
  – Decide:  
    a) keep filter and also sort by `CreatedDate DESC`, or  
    b) remove filter and rely purely on date sort.  
  – Consolidate logic in one method to avoid duplication.

────────────────────────────────────────────────────────
📦 4. Constants duplication (MED)  
────────────────────────────────────────────────────────
`SortingConstants` now live in both `Foundation.Catalog` and `Buyer.Web.Shared`. They can drift out of sync.  
RECOMMENDATION  
  • Extract to a shared project/package or generate one file from source.

────────────────────────────────────────────────────────
🖥️ 5. Front-end integration issues (MED)  
────────────────────────────────────────────────────────
• `FilterCollector` UI dispatches full `fetch` on every sort-selection without debounce; rapid clicks create race conditions.  
• Filtering drawer is merged but back-end filter support is *not* implemented (`// fetching a filtred data`).  
• Sort keys mismatch (`priceAsc`, `priceDsc` on FE vs BE constants).  
RECOMMENDATION  
  • Debounce or abort previous fetches.  
  • Hide filter UI until BE is ready.  
  • Align constants and write contract tests.
  
Addressing these will prevent performance regressions, maintain backward compatibility, and ensure stable paging behaviour.

dawiddworak88 (changes requested):
CRITICAL
1.  N² Elasticsearch updates  
    • In both InventoryProductsAvailableQuantityUpdateIntegrationEventHandler and OutletProductsAvailableQuantityUpdateIntegrationEventHandler the loop calls  
      ```await _productsService.BatchUpdateXxxQuantitiesAsync(@event.Products.Select(...))```  
      inside a foreach, resending the whole list each time (N×N writes).  
      ➜ Move the single batch call outside the loop.

2.  Inefficient + fragile ProductsService.BatchUpdate*  
    • Nested foreach (cultures × products) performs one UpdateAsync call per document.  
    • No bulk API, no retry → partial failure = silent data loss.  
      ➜ Build a BulkDescriptor per culture and call _elasticClient.BulkAsync; add retry/back-off.

3.  Duplicate, diverging SortingConstants  
    be/Foundation.Catalog.Definitions.SortingConstants and  
    fe/Buyer.Web/Shared/Definitions/Filters/SortingConstants use different casing values.  
      ➜ Keep one canonical source (shared package or API-exposed enum).

4.  Removed AppSettings class may break existing references  
    Catalog.Api’s AppSettings.cs is deleted but many pipelines/helm charts still rely on `ElasticsearchIndex`, `SupportedCultures`.  
      ➜ Search solution + deployment templates for “AppSettings”; migrate to typed options or IConfiguration.

HIGH
5.  Exception swallowing in Update*Quantity methods  
    Catch(Exception) → _logger.LogError(...); message ACKed, data lost.  
      ➜ Let exception propagate or implement DLQ/retry.

6.  O(n²) SKU lookup in BatchUpdate*  
    `FirstOrDefault` inside foreach.  
      ➜ Pre-materialise Dictionary<string,ProductSearchModel> by SKU.

7.  `SupportedCultures.Split(',')` without Trim → “ de ” yields bad ids.  
      ➜ `.Split(',').Select(c => c.Trim())`.

8.  Unvalidated `orderBy` query parameter flows directly to Elasticsearch sort builder (injection/DoS).  
      ➜ Validate against whitelist or enum.

MEDIUM
9.  Duplicate conditions in ProductSearchRepository.Sorting<T>; simplify logic.  
10. Nullable quantities added but many Razor/JS paths assume non-null ➜ potential NRE.  
11. FilterCollector: clearing filters doesn’t reset MUI Select internal state.  
12. TrackTotalHits() used on every search – unnecessary overhead.

PR #468, opened 2025-09-01
Add clickable slides in hero slider

PiotrZaranek:


dawiddworak88 (changes requested):
Findings and recommendations

1. HeroSlider.js: empty, self-closing anchor

- You added: {item.ctaUrl && <a href={item.ctaUrl} className="hero-slider__item__link" />}

- Issues:

  - Accessibility: An anchor with no text or aria-label does not have an accessible name. Screen readers will announce “link” with no context. This will fail common a11y linters/rules (“links must have discernible text”).
  - Semantics/SEO: A link with no content is poor semantics and could be interpreted as a generic link by assistive tech and crawlers.

- Fix:

  - Provide an accessible name (and optionally a title). Use available item fields.
  - If you intend to open links in new tabs, ensure rel="noopener noreferrer".
  - Example (minimal change to your current approach): {item.ctaUrl ? ( <a href={item.ctaUrl} className="hero-slider__item__link" aria-label={item.ctaText || item.teaserTitle || item.image?.imageAlt || "Open slide"} title={item.ctaText || item.teaserTitle || item.image?.imageTitle} {...(item.ctaTarget === "_blank" ? { target: "_blank", rel: "noopener noreferrer" } : {})} /> ) : null}

2. HeroSlider.scss: overlay not guaranteed to cover the slide

- Current: .hero-slider__item__link { display: block; width: 100%; height: 100%; position: absolute; }

- Issues:

  - Without top/left/right/bottom set (or inset), the absolutely positioned element may not cover the entire container. It will sit at its static position with width/height applied, which can fail to align exactly over the slide.
  - No z-index defined. This may result in the overlay appearing under other absolutely positioned elements (e.g., the teaser block), making only portions of the slide clickable.

- Fix:

  - Pin the overlay to all sides and set an appropriate z-index so it sits above slide content, but avoid interfering with carousel controls.

  - Provide visible focus treatment for keyboard users.

  - Example: .hero-slider { &__item { position: relative;

    ```javascript
    &__link {
      position: absolute;
      inset: 0;        // equivalent to top: 0; right: 0; bottom: 0; left: 0;
      display: block;
      z-index: 1;      // keep below carousel arrows if they’re outside the slide; adjust if necessary

      &:focus-visible {
        outline: 2px solid $primary;
        outline-offset: 2px;
      }
    }
    ```

    } }

3. Interaction with carousel swipe/controls

- Risk:

  - A full-bleed link can sometimes interfere with swipe gestures (touch) depending on how react-responsive-carousel handles pointer/click cancellation after drag. Usually it’s fine, but it’s worth testing across devices.
  - The link overlay should not overlap the carousel dots or arrows if those are outside the slide container, which appears to be the case. If you do see interference, verify stacking contexts and adjust z-index so arrows remain clickable on top.

4. Alternative (arguably better) markup

- Instead of an empty absolutely positioned link, wrap the slide’s contents in the anchor when ctaUrl is present. This improves semantics and accessibility without relying on an invisible overlay.

- HTML5 allows anchors to wrap block-level elements. Only avoid this if you plan to have nested interactive controls inside the slide (which would be invalid inside a link).

- Example: <div className="hero-slider__item"> {item.ctaUrl ? ( <a href={item.ctaUrl} className="hero-slider__item__link" // can simply be display:block, no absolute needed aria-label={item.ctaText || item.teaserTitle || item.image?.imageAlt || "Open slide"} title={item.ctaText || item.teaserTitle || item.image?.imageTitle} {...(item.ctaTarget === "_blank" ? { target: "_blank", rel: "noopener noreferrer" } : {})} > <ResponsiveImage {...item.image} /> {item.teaserTitle && ...} // teaser content </a> ) : ( <> <ResponsiveImage {...item.image} /> {item.teaserTitle && ...} </> )} </div>

  - Pros: Better accessibility, no empty link, no absolute positioning, and the entire content is semantically part of the link.
  - Cons: You cannot place additional interactive elements inside the slide when wrapped by a link.

5. Story data changes

- You replaced imageSrc URLs with a media-test domain; that’s fine for testing.
- Ensure that imageAlt/imageTitle are meaningful as they feed into the recommended aria-label/title fallbacks above.

6. Minor SCSS note on control dots

- You centered and limited width of control dots. Functionally unrelated to clickability, but acceptable. Ensure it doesn’t cause layout overlap with slides on smaller breakpoints.

Summary of concrete changes to meet the goal robustly

- In JS:
  - Add accessible naming to the anchor and optional target/rel handling.
- In SCSS:
  - Pin the overlay using inset: 0; and add z-index and focus-visible styles.
- Validate on touch devices to ensure swipe does not trigger unintended navigation.
- Consider the “wrap-all-in-anchor” alternative for best semantics if you don’t plan to add other interactive elements inside the slide.

Suggested patch snippet (minimal to your approach)

JS (inside .hero-slider__item): {item.ctaUrl ? ( <a href={item.ctaUrl} className="hero-slider__item__link" aria-label={item.ctaText || item.teaserTitle || item.image?.imageAlt || "Open slide"} title={item.ctaText || item.teaserTitle || item.image?.imageTitle} /> ) : null}

SCSS: .hero-slider { &__item { position: relative;

```javascript
&__link {
  position: absolute;
  inset: 0;
  display: block;
  z-index: 1;

  &:focus-visible {
    outline: 2px solid $primary;
    outline-offset: 2px;
  }
}
```

} }

With these adjustments, the implementation will reliably make the entire slide clickable, maintain keyboard navigability, and satisfy accessibility requirements.

dawiddworak88 (approved):

PR #469, opened 2025-09-01
Change clear cart button style

PiotrZaranek:


dawiddworak88 (changes requested):
Potential issues and recommendations

1. Missing color for destructive action

   - The clear button lost color="secondary" and now likely renders with default color, which may not communicate “destructive”.

   - Recommendation:

     - If using MUI v5: set color="error" variant="text" (or outlined) to communicate danger without competing visually.
     - If using MUI v4: keep color="secondary" only if your theme maps secondary to a destructive/red tone; otherwise create a theme override or a CSS “danger” class.
     - Consider size="small" to further lower prominence.

2. Import for Delete icon

   - startIcon={} requires an import. The diff doesn’t show it.
   - Ensure: import Delete from '@mui/icons-material/Delete' (MUI v5) or '@material-ui/icons/Delete' (MUI v4). Otherwise you’ll get a runtime/compile error.

3. Layout and responsiveness

   - Bulma is-flex is-justify-content-space-between pushes buttons to opposite sides on wide screens, which helps. On narrow screens, if both still fit one line, they’ll remain side-by-side and could still be scanned as a pair of actions.

   - Recommendations:

     - Stack on mobile:

       - Add a mobile column direction class (e.g., Bulma: is-flex-direction-column-mobile) or a custom class to switch to column, place clear below or above the primary.

       - Example class strategy:

         - Desktop: space-between, center aligned.
         - Mobile: column, full-width primary on top, clear as a left-aligned text button underneath.

     - Alternatively, move “Clear basket” out of the primary action area (e.g., under the basket list header or at the bottom-left under the items), which removes the “pairing” with Place order entirely.

4. Remove now-unused className

   - The button still has className="order__clear-button" but SCSS for that selector was removed. Either:

     - Remove the className from JSX, or
     - Reuse it for responsive rules (e.g., align-self or margin on mobile).

5. Accessibility and intent signaling

   - Add aria-label to clarify impact (e.g., aria-label="Clear all items from basket").
   - Ensure the clearBasket action triggers a confirmation flow. Your component shows a ConfirmationDialog for deletions — verify clearBasket indeed opens it (naming suggests it might perform the action directly).
   - Explicitly set type="button" on the clear button as a defensive measure inside forms (even though MUI defaults to "button").

6. Consistency with design system

   - Mixing Bulma layout utilities with MUI components is common but ensure consistent spacing units. Consider using a gap between buttons instead of relying only on space-between, to avoid edge cases when wrapping. For example, wrap in a container with gap via CSS and let clear button align-start on mobile.

7. Storybook and i18n coverage

   - You added clearBasketText to the story; make sure translations/defaults are available across locales.
   - Add a mobile viewport story to visually verify that the actions stack and the clear action remains clearly de-emphasized.

Suggested concrete tweak (illustrative)

- JSX:
  - Keep the flex container but add a custom class for responsive behavior and remove the unused class:
- SCSS: .order__actions { align-items: center; gap: 0.75rem; } @media (max-width: 768px) { .order__actions { flex-direction: column; align-items: stretch; } .order__actions button:last-child { align-self: flex-start; } }

Conclusion

- The change moves in the right direction by de-emphasizing and separating the “Clear basket” action. To fully meet the “not mistaken with place order” goal across viewports and for accessibility, explicitly mark it as destructive (color="error" or danger styling), ensure proper icon import, consider responsive stacking or relocating the clear action away from the primary action area, and clean up the unused class or repurpose it for responsive rules.

dawiddworak88 (approved):

PR #471, opened 2025-09-04
Add link to basket page in toast message

PiotrZaranek:


SebaKas21 (changes requested):
### 🚨 Must fix

1. **EN typo** – “beed” → “been” in `SuccessfullyAddedProduct`.
2. **Toast CTA** – make sure the “Go to basket” link comes from the backend VM, is culture-aware, and not hardcoded.
3. **Removed prop** – you dropped `SuccessfullyAddedProduct`; double-check the FE isn’t still using it or you’ll get runtime `undefined`.

---

### 👍 Nice to have

* Rename `ShowText` to something clearer like “View basket” in all locales.
* Handle multiple quick adds (stack/replace toasts).
* Add basic tests for the new VM (strings + basket URL).

---

### TL;DR

Fix the typo, feed the toast a proper basket URL from the VM, and make sure the old prop isn’t referenced anymore. After that, this PR is in good shape 🚀

SebaKas21 (approved):
### ⚠️ Please double-check / polish

1. **Basket URL source**
   Make sure the CTA’s href comes **from the backend VM** (culture-aware) and the FE uses client-side routing where applicable. If you’re still hardcoding `/basket`, switch to the VM value.

2. **CTA copy**
   If the key is still a generic `ShowText`, prefer something explicit like “View basket” in all locales. Better for a11y and analytics.

3. **Error path**
   Add an error toast for failed adds (with optional “Retry”). Make success/error visually distinct.

PR #473, opened 2025-09-05
Improving sliders by moving four elements by page

PiotrZaranek:


SebaKas21 (changes requested):
### 🚨 Things to fix

1. **Price component can crash**
   You removed the `&&` guard around `<Price {...carouselItem.price} />`. If `price` is ever missing, that’ll blow up or render garbage. Bring the check back or make `Price` handle empty props safely.

2. **Hard-coded `4` is risky**
   `slidesToSlide={4}` works fine on desktop, but on mobile it’ll feel broken (e.g. if only 1–2 cards are visible). Tie this to your responsive config instead of hardcoding.


---

### 👍 Strong suggestions

* Wrap this logic into a constant instead of having `4` inline. Something like `SLIDES_PER_PAGE_DESKTOP` in `CarouselConstants` keeps things consistent.
* Add prop-types (or TS) to catch stuff like missing `price`.
* Check behavior when there are fewer items than `slidesToSlide`. Some carousels render empty slides, some break navigation.

---

### Tests that would help

* Carousel with an item missing `price` → no crash.
* Mobile vs desktop carousel → step size matches visible items.

---

### Tiny nits

* Optional chaining is cleaner:

  ```jsx
  {carouselItem?.price && <Price {...carouselItem.price} />}
  ```
* Guard against empty `carouselItems` so you don’t render an empty titled block.

---

### TL;DR

Solid direction, but:

* put the `price` guard back,
* don’t hardcode `4`,

SebaKas21 (approved):
Double-check UX when items < slidesToSlide (esp. mobile). No action needed if the lib already handles it cleanly; otherwise cap the step.

PR #477, opened 2025-09-10
Add component to zoom images on mobile devices

PiotrZaranek:


dawiddworak88 (changes requested):
Here’s a focused code review of the diff for adding a mobile image zoom modal. I’ve grouped feedback by severity and provided concrete fixes.

Critical issues

1. Crash risk in ProductZoomModal due to destructuring before guards

- Problem: The code destructures mediaItems[index] before verifying mediaItems and index exist. This will throw if index is out of bounds or mediaItems is undefined.
- Fix:
  - Move the guard before destructuring and also ensure you only render for image items.

Example:

const ProductZoomModal = (props) => { const { isOpen, handleClose, mediaItems, index } = props;

if (!isOpen || !mediaItems || !mediaItems[index]) return null;

const item = mediaItems[index]; if (!item.mimeType?.startsWith('image')) return null;

const { mediaSrc, mediaAlt } = item; // ... };

2. Zoom modal opens for videos (and then tries to render a video URL as an )

- Problem: The onClick is attached to SplideSlide regardless of media type, but ProductZoomModal always renders an . If a video slide is clicked, zoom modal will still open and try to show video URL as an image.

- Fix:

  - Only open the zoom modal for items with mimeType starting with "image".
  - Apply the click handler only around the image content.

Example in ProductDetail:

{props.mediaItems.map((mediaItem, index) => ( {mediaItem.mimeType?.startsWith("image") ? ( <div onClick={() => handleOpenZoomModal(index)}> ) : ( )} ))}

3. Global CSS leakage affecting all MUI dialogs

- Problem: ProductZoomModal.scss styles .MuiDialog-paper globally. This will alter every dialog in the app whenever this SCSS is loaded (via ProductDetail.scss import).
- Fix:
  - Scope the style via PaperProps or classes so only this dialog is affected.

Example:

// JS

// SCSS .product-zoom-modal__paper { margin: 0.25rem !important; max-width: 700px !important; } .product-zoom-modal__image { width: 100%; height: 100%; object-fit: cover; transition: transform 200ms ease-out; touch-action: none; user-select: none; -webkit-user-drag: none; will-change: transform; }

High-priority improvements

4. Transform math and panning bounds are off

- Problem:

  - Current transform is scale(...) translate(...). In CSS, transforms are applied right-to-left; thus translate runs first, then scale scales the translation too. That makes the pan delta inconsistent and harder to constrain.
  - Constraining by image.width/10 is arbitrary and not tied to current scale. Users can drag beyond visible bounds or be restricted too much.

- Fix:

  - Use transform: translate(...) scale(...). This keeps translation in pixel space (easier to constrain).
  - Constrain panning based on current scale and the image rect.

Example:

// style style={{ transform: `translate(${position.x}px, ${position.y}px) scale(${scale})`, transformOrigin: "center center", }}

// constrainPosition const constrainPosition = (pos) => { const rect = imageRef.current?.getBoundingClientRect(); if (!rect) return pos;

if (scale <= 1) return ZoomModalConstants.defaultPosition();

const maxX = (rect.width * (scale - 1)) / 2; const maxY = (rect.height * (scale - 1)) / 2;

return { x: Math.max(-maxX, Math.min(maxX, pos.x)), y: Math.max(-maxY, Math.min(maxY, pos.y)), }; };

5. Double-tap centering should match new transform order

- Problem: The position calculation assumes the previous transform order and divides by scale, which doesn’t align with translate-then-scale.
- Fix:

const handleDoubleTap = (touch) => { if (scale > 1) { resetZoom(); setIsZoomed(false); return; } const rect = imageRef.current.getBoundingClientRect(); const dx = touch.clientX - (rect.left + rect.width / 2); const dy = touch.clientY - (rect.top + rect.height / 2);

const next = ZoomModalConstants.defaultZoom(); setScale(next); setIsZoomed(true); setPosition(constrainPosition({ x: -dx * (next - 1), y: -dy * (next - 1), })); };

6. Guard the image/video checks everywhere

- Problem: ProductDetail uses mediaItem.mimeType.startsWith("image"). If mimeType is undefined, this will throw.
- Fix: Use optional chaining: mediaItem.mimeType?.startsWith("image")

7. Mobile-only behavior gating

- Requirement states “zooming in images on mobile devices,” but the new modal opens on any click/tap (including desktop).
- Improvement: Gate by touch capability or viewport.

Examples:

- Only attach the click handler if navigator.maxTouchPoints > 0
- Or in the modal, early return when not on touch devices

const isTouch = typeof navigator !== "undefined" && navigator.maxTouchPoints > 0; // ... {isTouch && mediaItem.mimeType?.startsWith("image") && (

Nice-to-have enhancements

8. Pinch-to-zoom

- Currently multi-touch is blocked. Consider adding pinch gesture to adjust scale between min/max and recenter as needed.

9. Fullscreen on small screens

- Consider making the dialog fullScreen on small breakpoints for a cleaner UX (MUI’s useMediaQuery).

10. Naming and PropTypes

- Consider renaming lastPositionref -> lastPositionRef for consistency.
- Add PropTypes to ProductZoomModal.

Example:

import PropTypes from "prop-types"; ProductZoomModal.propTypes = { isOpen: PropTypes.bool.isRequired, handleClose: PropTypes.func.isRequired, mediaItems: PropTypes.arrayOf(PropTypes.shape({ mediaSrc: PropTypes.string.isRequired, mediaAlt: PropTypes.string, mimeType: PropTypes.string, })).isRequired, index: PropTypes.number.isRequired, };

11. Storybook data fidelity

- mimeType in stories is "image" rather than a real value like "image/jpeg". startsWith("image") still works, but it’s better to simulate actual API responses.
- Ensure stories for video use realistic mime types like "video/mp4" and verify the zoom doesn’t open on them.

Test checklist

- Clicking image slides on touch devices opens the zoom modal; clicking video slides does not.
- No runtime errors when mediaItems is empty or index is out of range.
- Double-tap to zoom in focuses around the tapped point; double-tap again resets.
- Panning is smooth and constrained (no empty gutter showing beyond image edges).
- Dialog styling only affects this modal (no changes to other dialogs).
- On desktop (mouse), clicking doesn’t open zoom if it’s intended to be mobile-only.
- Backdrop click and ESC close the modal and reset zoom state.
- Orientation changes don’t break layout; image fills dialog as expected.

Summary

- Must-fix: destructuring before guards, opening for videos, global dialog CSS leakage.
- Strongly recommended: adjust transform order and pan constraints, add optional-chaining for mimeType, gate to mobile interactions.
- Optional: pinch-to-zoom, fullscreen on small screens, PropTypes, naming polish, story fidelity.

Applying these changes will make the zoom modal robust, scoped, and aligned with the stated goal of mobile image zoom while preventing regressions elsewhere in the app.

PR #487, opened 2025-09-26
Add getting price in OrderFileApiController

PiotrZaranek:
additional works:
- Added GetProductsBySkusAsync to ProductsRepository in Buyer.Web
- Added client to onDrop useCallback function to send id to API

### DIFF
--- a/be/src/Project/Web/Buyer/Buyer.Web/Areas/Orders/ApiControllers/OrderFileApiController.cs
+++ b/be/src/Project/Web/Buyer/Buyer.Web/Areas/Orders/ApiControllers/OrderFileApiController.cs
@@ -36,7 +36,6 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Net;
-using System.Text.Json;
 using System.Threading.Tasks;
 
 namespace Buyer.Web.Areas.Orders.ApiControllers
@@ -109,9 +108,7 @@ namespace Buyer.Web.Areas.Orders.ApiControllers
                 return StatusCode((int)HttpStatusCode.BadRequest, new { Message = _orderLocalizer.GetString("ProductsNotFound") });
             }
 
-            var productBySku = products
-                .OrEmptyIfNull()
-                .ToDictionary(g => g.Sku, g => g);
+            var orderLines = importedOrderLines.Where(x => products.Select(y => y.Sku).Contains(x.Sku));
 
             var productIds = products.OrEmptyIfNull().Select(x => x.Id).Distinct();
             var stockAvailableProducts = await _inventoryRepository.GetStockAvailbleProductsByProductIdsAsync(token, language, productIds);
@@ -126,13 +123,13 @@ namespace Buyer.Web.Areas.Orders.ApiControllers
             {
                 var priceProducts = new List<PriceProduct>();
 
-                foreach (var orderLine in importedOrderLines)
+                foreach (var orderLine in orderLines)
                 {
                     var product = products.FirstOrDefault(x => x.Sku == orderLine.Sku);
 
                     if (product is null)
                     {
-                        priceProducts.Add(null);
+                        continue;
                     }
 
                     priceProducts.Add(new PriceProduct
@@ -171,9 +168,9 @@ namespace Buyer.Web.Areas.Orders.ApiControllers
                     });
             }
 
-            for (var i = 0; i < importedOrderLines.Count(); i++)
+            for (var i = 0; i < orderLines.Count(); i++)
             {
-                var orderLine = importedOrderLines.ElementAtOrDefault(i);
+                var orderLine = orderLines.ElementAtOrDefault(i);
 
                 if (orderLine is null)
                 {
--- a/be/src/Project/Web/Buyer/Buyer.Web/Areas/Orders/ApiControllers/OrderFileApiController.cs
+++ b/be/src/Project/Web/Buyer/Buyer.Web/Areas/Orders/ApiControllers/OrderFileApiController.cs
@@ -168,7 +168,7 @@ namespace Buyer.Web.Areas.Orders.ApiControllers
                     continue;
                 }
 
-                var availableStock = stockByProductId.GetValueOrDefault(product.Id);
+                var availableStock = stockByProductId.GetValueOrDefault(product.Id) ?? 0;
 
                 var stockQuantity = Math.Min(orderLine.Quantity, (double)availableStock);
                 var quantity = orderLine.Quantity - stockQuantity;
@@ -177,9 +177,6 @@ namespace Buyer.Web.Areas.Orders.ApiControllers
                 var pictureUrl = firstImage != Guid.Empty
                     ? _mediaService.GetMediaUrl(firstImage, OrdersConstants.Basket.BasketProductImageMaxWidth)
                     : null;
-
-                var unitPrice = price?.CurrentPrice ?? 0m;
-                var totalPrice = unitPrice * (decimal)orderLine.Quantity;
 
                 var basketItem = new BasketItem
                 {
@@ -190,11 +187,17 @@ namespace Buyer.Web.Areas.Orders.ApiControllers
                     Quantity = quantity,
                     StockQuantity = stockQuantity,
                     ExternalReference = orderLine.ExternalReference,
-                    MoreInfo = orderLine.MoreInfo,
-                    UnitPrice = unitPrice,
-                    Price = totalPrice,
-                    Currency = price?.CurrencyCode ?? string.Empty,
+                    MoreInfo = orderLine.MoreInfo
                 };
+
+                if (price is not null)
+                {
+                    var totalPrice = price.CurrentPrice * (decimal)orderLine.Quantity;
+
+                    basketItem.UnitPrice = price.CurrentPrice;
+                    basketItem.Price = totalPrice;
+                    basketItem.Currency = price.CurrencyCode;
+                }
 
                 basketItems.Add(basketItem);
             }
--- a/be/src/Project/Web/Buyer/Buyer.Web/Areas/Orders/ApiControllers/OrderFileApiController.cs
+++ b/be/src/Project/Web/Buyer/Buyer.Web/Areas/Orders/ApiControllers/OrderFileApiController.cs
@@ -158,17 +158,18 @@ namespace Buyer.Web.Areas.Orders.ApiControllers
                     });
             }
 
-            var linesWithPrices = importedOrderLines.OrEmptyIfNull().Zip(prices);
+            var productsWithPrices = products.OrEmptyIfNull().Zip(prices);
 
-            foreach (var (orderLine, price) in linesWithPrices)
+            foreach (var (product, price) in productsWithPrices)
             {
-                if (!productBySku.TryGetValue(orderLine.Sku, out var product) || product == null)
+                var availableStock = stockByProductId.GetValueOrDefault(product.Id) ?? 0;
+
+                var orderLine = importedOrderLines.FirstOrDefault(x => x.Sku == product.Sku);
+
+                if (orderLine is null)
                 {
-                    _logger.LogError($"Product for SKU {orderLine.Sku} and language {language} couldn't be found.");
                     continue;
                 }
-
-                var availableStock = stockByProductId.GetValueOrDefault(product.Id) ?? 0;
 
                 var stockQuantity = Math.Min(orderLine.Quantity, (double)availableStock);
                 var quantity = orderLine.Quantity - stockQuantity;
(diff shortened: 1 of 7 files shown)


dawiddworak88 (changes requested):
It's important to check the order items don't disappear when the price is missing. It should be an order item line added with simply empty price.

Goal: add prices to order items when uploading via drag-and-drop Excel, but still work (import lines) when no price is available by leaving price fields empty.

Key findings

Critical functional regressions
- Items silently disappear when prices are not available (both Buyer and Seller)
  - Cause: using products.Zip(prices) to drive the loop. If prices is empty (e.g., feature disabled, token missing, service down) or counts/orders don’t match, the loop produces zero pairs, so no BasketItem is added at all.
  - Requirement violation: items should still be imported with empty UnitPrice/Price/Currency.

- Potential mispricing due to misalignment between products and prices
  - Cause: Zip pairs by index, not by identity. If the pricing service returns prices in a different order, or a subset/superset, prices will be associated with wrong products or skipped.
  - Correct approach: associate by a stable key (PrimarySku and/or Sku).

- Duplicate SKU rows from the Excel file are lost
  - Cause: iteration shifted from “foreach orderLine” to “foreach (product, price) from Zip” and then FirstOrDefault(x => x.Sku == product.Sku). If Excel contains multiple rows for the same SKU (with different qty or metadata), all but the first are ignored.

Correctness/compile-time issues
- availableStock retrieval now uses GetValueOrDefault(...) ?? 0 on a Dictionary<Guid, T>. GetValueOrDefault returns a non-nullable value type, so the null-coalescing ?? 0 is invalid (and will not compile unless the value is nullable). The previous TryGetValue fallback was correct.

Resilience/robustness issues
- Buyer: Guid.Parse on the client claim Id can throw if present but malformed. Prefer Guid.TryParse.
- Attribute helpers: .ToYesOrNo() and ProductColorsService.ToEnglishAsync(...) are called on possibly-null attribute values. Ensure extension methods can handle null safely.
- Pricing fetch: not wrapped in try/catch; a transient failure will currently blow up the request. Requirement says “should still work when there is no price,” so fall back to no-price gracefully.

Behavior/parity issues
- Buyer: pricing fetch gated by _options.Value.GrulaAccessToken—which is fine. But because of Zip, when the gate prevents pricing, no items are added. Needs the refactor below.
- Seller: also gates by GrulaAccessToken while calling a IPriceService overload without passing the token. This might be OK if Seller PriceService reads the token from configuration; however, it’s inconsistent with Buyer and confusing. Consider aligning these patterns.

Minor issues and suggestions
- Logging: in the old code, missing SKU lines logged an error per line. New code no longer logs when some order lines have unmatched SKUs; recommend logging a warning per missing line for traceability.
- Image selection: your FirstOrDefault + Guid.Empty check is fine and avoids exceptions; it’s an improvement over First() + Any().
- ProductsRepository.GetProductsBySkusAsync:
  - Pagination loop: when throwing inside the pagination loop, you throw with response.Message instead of nextPagesResponse.Message; this loses the real error context.
  - Consider returning Enumerable.Empty<Product>() instead of default to avoid null downstream.

Concrete fixes

1) Replace the Zip-based loop with an order-line-driven loop keyed by SKU, and map prices by a stable key.

Buyer.Web Areas Orders OrderFileApiController.Index

- Keep product lookup and stock retrieval as before, but iterate order lines to preserve duplicates.
- Build a dictionary of prices keyed by PrimarySku (and fall back to Sku if appropriate).
- If a price is found, set UnitPrice/Price/Currency; otherwise leave them null.
- Keep robust fallbacks when pricing is disabled or fails.

Illustrative code sketch (Buyer)

var importedOrderLines = _orderFileService.ImportOrderLines(model.File);
var basketItems = new List<BasketItem>();

var token = await HttpContext.GetTokenAsync(ApiExtensionsConstants.TokenName);
var language = CultureInfo.CurrentUICulture.Name;

var skus = importedOrderLines.OrEmptyIfNull().Select(x => x.Sku).Distinct();
var products = await _productsRepository.GetProductsBySkusAsync(token, language, skus);

if (products.OrEmptyIfNull().Any() is false)
{
    return StatusCode((int)HttpStatusCode.BadRequest, new { Message = _orderLocalizer["ProductsNotFound"] });
}

var productBySku = products.OrEmptyIfNull().ToDictionary(p => p.Sku, p => p, StringComparer.OrdinalIgnoreCase);

var productIds = products.OrEmptyIfNull().Select(x => x.Id).Distinct();
var stockAvailableProducts = await _inventoryRepository.GetStockAvailbleProductsByProductIdsAsync(token, language, productIds);
var stockByProductId = stockAvailableProducts.OrEmptyIfNull().ToDictionary(s => s.ProductId, s => s.AvailableQuantity);

// Build price dictionary (safe, optional)
Dictionary<string, Price> priceByPrimarySku = new(StringComparer.OrdinalIgnoreCase);

try
{
    if (string.IsNullOrWhiteSpace(_options.Value.GrulaAccessToken) is false)
    {
        var priceProductsTasks = products.Select(async x => new PriceProduct
        {
            PrimarySku = x.PrimaryProductSku,
            FabricsGroup = _productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossiblePriceGroupAttributeKeys),
            ExtraPacking = _productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossibleExtraPackingAttributeKeys).ToYesOrNo(),
            SleepAreaSize = _productsService.GetSleepAreaSize(x.ProductAttributes),
            PaletteSize = _productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossiblePaletteSizeAttributeKeys),
            Size = _productsService.GetSize(x.ProductAttributes),
            PointsOfLight = _productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossiblePointsOfLightAttributeKeys),
            LampshadeType = _productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossibleLampshadeTypeAttributeKeys),
            LampshadeSize = _productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossibleLampshadeSizeAttributeKeys),
            LinearLight = _productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossibleLinearLightAttributeKeys).ToYesOrNo(),
            Mirror = _productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossibleMirrorAttributeKeys).ToYesOrNo(),
            Shape = _productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossibleShapeAttributeKeys),
            PrimaryColor = await _productColorsService.ToEnglishAsync(_productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossiblePrimaryColorAttributeKeys)),
            SecondaryColor = await _productColorsService.ToEnglishAsync(_productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossibleSecondaryColorAttributeKeys)),
            ShelfType = _productsService.GetFirstAvailableAttributeValue(x.ProductAttributes, _options.Value.PossibleShelfTypeAttributeKeys)
        });

        var priceProducts = await Task.WhenAll(priceProductsTasks);

        Guid? clientId = null;
        var clientIdClaim = User.FindFirst(ClaimsEnrichmentConstants.ClientIdClaimType)?.Value;
        if (Guid.TryParse(clientIdClaim, out var parsedClientId))
            clientId = parsedClientId;

        var priceClient = new PriceClient
        {
            Id = clientId,
            Name = User.Identity?.Name,
            CurrencyCode = User.FindFirst(ClaimsEnrichmentConstants.CurrencyClaimType)?.Value,
            ExtraPacking = User.FindFirst(ClaimsEnrichmentConstants.ExtraPackingClaimType)?.Value,
            PaletteLoading = User.FindFirst(ClaimsEnrichmentConstants.PaletteLoadingClaimType)?.Value,
            Country = User.FindFirst(ClaimsEnrichmentConstants.CountryClaimType)?.Value,
            DeliveryZipCode = User.FindFirst(ClaimsEnrichmentConstants.ZipCodeClaimType)?.Value
        };

        var prices = await _priceService.GetPrices(_options.Value.GrulaAccessToken, DateTime.UtcNow, priceProducts, priceClient);

        // Assuming Price has PrimarySku; if not, adjust mapping to the available identity key.
        priceByPrimarySku = prices.OrEmptyIfNull()
                                  .Where(p => !string.IsNullOrWhiteSpace(p.PrimarySku))
                                  .GroupBy(p => p.PrimarySku, StringComparer.OrdinalIgnoreCase)
                                  .ToDictionary(g => g.Key, g => g.First(), StringComparer.OrdinalIgnoreCase);
    }
}
catch (Exception ex)
{
    _logger.LogWarning(ex, "Price service failed; proceeding without prices.");
}

foreach (var orderLine in importedOrderLines.OrEmptyIfNull())
{
    if (!productBySku.TryGetValue(orderLine.Sku, out var product) || product is null)
    {
        _logger.LogWarning("Product for SKU {Sku} and language {Language} not found.", orderLine.Sku, language);
        continue;
    }

    var availableStock = stockByProductId.TryGetValue(product.Id, out var qty) ? qty : 0;
    var stockQuantity = Math.Min(orderLine.Quantity, (double)availableStock);
    var quantity = orderLine.Quantity - stockQuantity;

    var firstImage = product.Images.OrEmptyIfNull().FirstOrDefault();
    var pictureUrl = firstImage != Guid.Empty ? _mediaService.GetMediaUrl(firstImage, OrdersConstants.Basket.BasketProductImageMaxWidth) : null;

    var basketItem = new BasketItem
    {
        ProductId = product.Id,
        ProductSku = product.Sku,
        ProductName = product.Name,
        PictureUrl = pictureUrl,
        Quantity = quantity,
        StockQuantity = stockQuantity,
        ExternalReference = orderLine.ExternalReference,
        MoreInfo = orderLine.MoreInfo
    };

    if (product.PrimaryProductSku is not null &&
        priceByPrimarySku.TryGetValue(product.PrimaryProductSku, out var price) &&
        price is not null)
    {
        basketItem.UnitPrice = price.CurrentPrice;
        basketItem.Price = price.CurrentPrice * (decimal)orderLine.Quantity;
        basketItem.Currency = price.CurrencyCode;
    }

    basketItems.Add(basketItem);
}

- Apply the same pattern in Seller’s OrderFileApiController (Index): build price dictionary once; iterate importedOrderLines; attach price if found; otherwise leave empty. Also wrap GetPrices in try/catch and log a warning, then continue without prices.

2) Fix available stock retrieval

Replace:
var availableStock = stockByProductId.GetValueOrDefault(product.Id) ?? 0;

With:
var availableStock = stockByProductId.TryGetValue(product.Id, out var qty) ? qty : 0;

3) Robust Guid parsing (Buyer)

Replace Guid.Parse(...) with Guid.TryParse as in the sketch above to avoid exceptions on malformed claims.

4) Null-safe helpers

Ensure ToYesOrNo() safely handles null/empty values. Likewise, make sure ProductColorsService.ToEnglishAsync gracefully returns null when given null input. If not, guard with a null check before calling.

5) Logging

Restore per-line logging for missing SKUs:
_logger.LogWarning("Product for SKU {Sku} and language {Language} not found.", orderLine.Sku, language);

6) Align Seller pricing enablement

If Seller’s IPriceService resolves the access token from configuration, the GrulaAccessToken gate is OK. Otherwise consider:
- If a token is required, pass it (mirror Buyer signature), or
- If not required, remove the gating on GrulaAccessToken to avoid skipping prices accidentally.

Either way, the code should still import lines when pricing cannot be computed (handled by the loop refactor).

7) ProductsRepository.GetProductsBySkusAsync fixes

- In pagination error, throw the correct response object:
throw new CustomException(nextPagesResponse.Message, (int)nextPagesResponse.StatusCode);

- At the end, return Enumerable.Empty<Product>() instead of default.

Testing recommendations

- No-price path: configure blank GrulaAccessToken and verify that items are created with UnitPrice/Price/Currency null, not dropped.
- Partial-price path: mock pricing to return a subset; verify only the matching items get a price, others remain empty.
- Reordered pricing: mock pricing returning items in a different order; verify mapping by PrimarySku is correct.
- Duplicate SKU lines in Excel: two rows with same SKU but different quantities; verify both lines are imported separately and each priced correctly.
- Missing SKU(s): simulate missing product(s); verify the successful ones import and warnings are logged for missing ones.
- Malformed client claim (Buyer): ensure no exception and items still added with empty price.

Conclusion

The feature direction is correct, but the Zip-based loop introduces severe regressions in core behavior. Refactoring to an order-line-driven loop with price lookup by a stable key, plus the stock retrieval and robustness fixes above, will meet the requirement: import all items and attach prices where available, leaving the fields empty otherwise.

SebaKas21 (approved):

dawiddworak88 (approved):

PR #499, opened 2025-10-06
Added filters for backend

SebaKas21:


SebaKas21:
@dawiddworak88 In the first step, I would like to make code review fixes and then test them in  test environment with changing the index, etc.

PiotrZaranek (changes requested):
## 🔴 Critical Issues

### 1. **Typo in Parameter Names**
Throughout the code, `language` is misspelled as `langauge`:
```csharp
public async Task<PagedResultsWithFilters<IEnumerable<ProductSearchModel>>> GetPagedResultsWithFilters(
    string langauge,  // ❌ Should be "language"
```
**Impact:** This is inconsistent with the rest of the codebase and reduces code readability.

### 2. **Incorrect OR Logic in PostFilter**
```csharp
.PostFilter(fq =>
    fq.Terms(t => t.Field("categoryName.keyword").Terms(filters.Category)) ||
    fq.Nested(n => n.Path("productAttributes")...
```
**Problem:** Using `||` (OR) means products matching EITHER category OR attributes will be returned. This is likely incorrect - you probably want products that match ALL selected filters (AND logic).

**Fix:** Use `&&` and wrap each filter group properly, or better yet, build filters conditionally:
```csharp
.PostFilter(fq => {
    var query = fq.Bool(b => {
        var boolQuery = b;
        
        if (filters.Category?.Any() == true)
            boolQuery = boolQuery.Filter(f => f.Terms(t => t.Field("categoryName.keyword").Terms(filters.Category)));
        
        if (filters.Color?.Any() == true || filters.Shape?.Any() == true || ...)
            boolQuery = boolQuery.Filter(f => f.Nested(...));
        
        return boolQuery;
    });
    return query;
})
```

### 3. **Null Reference Risk**
```csharp
if (response.IsValid)
{
    return new PagedResultsWithFilters<IEnumerable<ProductSearchModel>>(response.Total, itemsPerPage.Value)
```
`itemsPerPage.Value` could throw if `itemsPerPage` is null (even though you set defaults earlier, it's safer to be explicit).

### 4. **Empty Filter Arrays Not Handled**
The code doesn't check if filter arrays are null or empty before adding them to queries:
```csharp
fq.Terms(t => t.Field("categoryName.keyword").Terms(filters.Category))
```
If `filters.Category` is null or empty, this will create an invalid query.

---

## ⚠️ Major Issues

### 5. **Code Duplication**
The two `GetPagedResultsWithFilters` methods have ~90% identical code. The query building and aggregation logic is duplicated.

**Recommendation:** Extract common logic:
```csharp
private async Task<PagedResultsWithFilters<IEnumerable<ProductSearchModel>>> ExecuteSearchWithFilters(
    QueryContainer baseQuery,
    QueryFilters filters,
    int pageIndex,
    int itemsPerPage,
    string orderBy)
{
    // Common search logic here
}
```

### 6. **Hardcoded Range Values**
Range aggregations have hardcoded values:
```csharp
.Range("height", r => r.Field("productAttributes.height.value")
    .Ranges(
        rr => rr.From(0).To(69),
        rr => rr.From(70).To(79),
        // ...
    )
)
```
**Problem:** These ranges are domain-specific and should be:
- Extracted to configuration
- Or made dynamic based on actual data distribution
- Or at minimum, moved to constants

### 7. **Nested Filter Logic Issue**
```csharp
.Filter(
    fq => fq.Terms(...Color),
    fq => fq.Terms(...Shape),
    fq => nq.Bool(b2 => b2.Should(BuildRangeQueries(filters.Height, ...)...)
```
All filters are in `.Filter()` which means AND logic, but range queries use `.Should()` with `MinimumShouldMatch(1)`. This means:
- Color AND Shape AND (any height range) AND (any width range) AND (any depth range)

This might be correct, but seems unusual. Typically you'd want: "match any of the selected height ranges" which is what you have, but the overall structure is confusing.

### 8. **Range Query Parsing Logic**
```csharp
private IEnumerable<QueryContainer> BuildRangeQueries(IEnumerable<string>? ranges, string field)
{
    // ...
    lt = to + 1;  // ❌ Why +1?
```
The `+1` logic is inconsistent with how ranges are formatted in `MapBucketAggregate`:
```csharp
$"{from:0}-{(to - 1):0}";  // Formats as "0-68"
```
But then you parse it and add `+1` back. This is confusing and error-prone.

**Better approach:** Be consistent - either ranges are inclusive (0-69) or exclusive (0-70).

---

## 💡 Moderate Issues

### 9. **Missing Null Checks**
```csharp
public class QueryFilters
{
    public string[] Category { get; set; }
    // ...
}
```
Properties should be nullable or initialized to empty arrays:
```csharp
public string[]? Category { get; set; }
// OR
public string[] Category { get; set; } = Array.Empty<string>();
```

### 10. **Inconsistent ItemsPerPage Handling**
In the second overload:
```csharp
return new PagedResultsWithFilters<IEnumerable<ProductSearchModel>>(response.Total, (int)response.Total)
```
Using `response.Total` as page size is wrong - it should use a sensible default or the actual items returned.

### 11. **No Validation on QueryFilters**
The controller accepts filters without validation. Consider:
- Maximum array lengths (prevent DOS attacks)
- Valid range formats
- Sanitization of input values

### 12. **Inefficient IDs Query Building**
```csharp
var idsQuery = Query<ProductSearchModel>.MatchNone();
foreach (var id in ids)
{
    idsQuery = idsQuery || Query<ProductSearchModel>.Term(t => t.Field(x => x.ProductId).Value(id));
}
```
This is inefficient. Use `.Terms()` instead:
```csharp
var idsQuery = Query<ProductSearchModel>.Terms(t => t.Field(x => x.ProductId).Terms(ids));
```

### 13. **Aggregation Size Limit**
```csharp
.Terms("category", t => t.Field("categoryName.keyword").Size(100))
```
Hardcoded to 100. What if there are more categories? Consider:
- Making this configurable
- Using composite aggregations for large cardinality fields
- At minimum, document this limitation

---

## 🟡 Minor Issues

### 14. **Controller Returns `default` Instead of NotFound**
```csharp
return default;
```
Should return a proper HTTP status:
```csharp
return NotFound();
```

### 15. **Commented Code in Repository Interface**
```csharp
//Task<Paged<IEnumerable<Product>> GetProductsWithFiltersAsync(string token, string language, string searchTerm,)
```
Remove commented-out code before merging.

### 16. **Inconsistent Naming**
- `pageResullt` - typo, should be `pageResult`
```csharp
var pageResullt = new PagedResults<IEnumerable<ProductSearchModel>>(...)
```

### 17. **Model Properties Without Descriptions**
```csharp
public class Filter
{
    public string Name { get; set; }
    public List<string> Values { get; set; }
}
```
Add XML documentation comments for API models.

### 18. **Docker Compose Index Change**
```yaml
- ElasticsearchIndex=catalog2
```
This looks like a temporary change for testing. Should this be reverted before merging?

---

## 🎯 Recommendations

### Performance Optimizations
1. **Cache Aggregations:** Filter options likely don't change frequently - consider caching
2. **Selective Aggregations:** Only return aggregations that have results
3. **Index Optimization:** Ensure fields used in filters are properly mapped with `keyword` type

### Code Quality
1. Extract magic numbers to constants
2. Add comprehensive unit tests for `BuildRangeQueries` and `MapBucketAggregate`
3. Add integration tests for filter combinations
4. Consider using a query builder pattern to reduce complexity

### Security
1. Add input validation on filter values
2. Consider rate limiting on search endpoints
3. Sanitize user inputs before constructing queries

---

## ✅ Positive Aspects
- Good separation of concerns (repository → service → controller)
- Proper use of async/await
- Aggregations implementation is correct conceptually
- Returns filters alongside results (good UX)

---

## Summary
The PR implements the core functionality but needs refinement before merging. Focus on:
1. **Fix the typo** (`langauge` → `language`)
2. **Fix the filter logic** (OR vs AND)
3. **Remove code duplication**
4. **Handle null/empty filters properly**
5. **Fix the IDs query performance issue**
6. **Clean up temporary changes** (elasticsearch index name, commented code)

dawiddworak88 (approved):
We will review add-filters-component PR to develop

PR #502, opened 2025-10-20
Orders filtering via order statuses

SebaKas21:


dawiddworak88 (changes requested):
W'e're almost there. Small suggested changes summary :) :

- FE bug: Fix PropTypes in OrdersStatusFilters: selectedStatusId: PropTypes.oneOfType([PropTypes.string, PropTypes.oneOf([null])])

- FE cleanup: Remove unused imports in Catalog.js.

- Optional BE enhancement: If desired, support orderStatusId in the “ids” branch too.

- i18n: Consider Polish “Wszystkie” instead of “Wszystko”.

dawiddworak88 (approved):

PR #507, opened 2025-12-03
Resolved files uploading issue

SebaKas21:


dawiddworak88 (changes requested):
Here’s a structured code review of the diff, with remarks ranked by importance (highest first) and focused on whether the change really fixes file upload and what risks it introduces.

---
## 1. CRITICAL – `uploadId` is only generated for the **first** file

**Location (FE):** `MediaCloud.js`

```js
const [uploadId, setUploadId] = useState(null);
...
const filesInChunksUpload = () => {
    if (fileToUploadInChunks) {        
        if (chunkCounter < chunkCount) {
            const chunk = fileToUploadInChunks.slice(beginingOfTheChunk, endOfTheChunk);

            uploadChunk(chunk);
        }
    }
    else {
        const file = filesToUploadInChunks[fileToUploadInChunksIndex];

        const newUploadId = uuidv4();

        setUploadId(newUploadId);
        ...
        setFileToUploadInChunks(file);
    }
};
...
if ((fileToUploadInChunksIndex + 1) < filesToUploadInChunks.length) {
    setFileToUploadinChunksIndex(fileToUploadInChunksIndex + 1);
    setFileToUploadInChunks(filesToUploadInChunks[fileToUploadInChunksIndex + 1]);
    ...
    setUploadId(null);
}
```

- For the **first** file, `fileToUploadInChunks` is `null`, so you enter the `else` branch, generate `newUploadId`, and set it.
- For the **second and subsequent** files, you set `fileToUploadInChunks` directly and **never** go through the `else` branch again; `uploadId` is reset to `null` and never regenerated.
- `uploadChunk` always sends whatever is in `uploadId`:

  ```js
  formData.append("uploadId", uploadId);
  ```

- On the server side, `UploadId` is required by validators (`NotNull().NotEmpty()`), and it’s also used in the filename:

  ```csharp
  var fileName = $"{model.UploadId}_{model.ChunkSumber}";
  ```

  If `UploadId` is null, this is a guaranteed problem (either validation failure upstream or NRE when using it).

**Impact:**
- Multi-file upload in a single operation will most likely break after the first file.
- This undermines the primary goal of the PR (reliable file upload).

**Recommendation:**
- Treat `uploadId` as a **per-file** value (e.g., part of the file metadata array) instead of a single global state.
- When switching to a new file (`fileToUploadInChunksIndex + 1`), generate a new `uploadId` there before you start chunking that file.

---
## 2. CRITICAL – Inconsistent `UploadId` nullability and potential runtime issues

Across layers, `UploadId` is sometimes nullable, sometimes not:

- **Media API service models**
  - `CreateFileChunkServiceModel.UploadId` – `Guid?`
  - `CreateMediaItemFromChunksServiceModel.UploadId` – `Guid?`
  - `UpdateMediaItemFromChunksServiceModel.UploadId` – `Guid?`
- **Media API request models**
  - `UploadMediaChunkRequestModel.UploadId` – `Guid?`
  - `UploadMediaChunksCompleteRequestModel.UploadId` – `Guid?`
- **ApiExtensions models**
  - `FileRequestModelBase.UploadId` – `string`
  - `FileChunksSaveCompleteRequestModel.UploadId` – `Guid` (non-nullable)
- **Seller.Web**
  - `UploadMediaChunkCompleteRequestModel.UploadId` – `Guid` (non-nullable)
  - `FilesRepository.SaveChunksCompleteAsync(..., Guid uploadId)`

On the Media side, validators require `UploadId`:

```csharp
this.RuleFor(x => x.UploadId).NotNull().NotEmpty();
```

But then in `MediaService` you do:

```csharp
var fileName = $"{model.UploadId}_{model.ChunkSumber}"; // model.UploadId is Guid?
```

If validation is missed or there’s any path that bypasses it, `model.UploadId` can be null and this will blow up at runtime.

**Impact:**
- Hard runtime failures if validation is not consistently applied before service calls.
- Inconsistent type usage makes the code harder to reason about and increases chances of subtle bugs.

**Recommendation:**
- Decide if `UploadId` is **required** (it looks like it is) and then:
  - Make it **non-nullable** (`Guid`) in all server-side models that require it, especially service models.
  - Or, if you really need `Guid?` in models, never dereference it without a `HasValue` check and early fail with a clear error.
- Align ApiExtensions / Seller.Web / Media.Api on the same conceptual type (see remark 4).

---
## 3. HIGH – Breaking changes for existing clients & error handling

**Breaking behavior introduced:**

- Media API validators now require `UploadId` for:
  - `CreateFileChunkServiceModel` (via `CreateFileChunkModelValidator`)
  - `CreateMediaItemFromChunksServiceModel`
  - `UpdateMediaItemFromChunksServiceModel`
- Seller.Web now calls `SaveChunkAsync` and `SaveChunksCompleteAsync` with `uploadId` and `UploadId` parameters that are effectively mandatory for success.

Any client still using the old contract (no `uploadId` / `UploadId`) will:

- Successfully hit `Seller.Web`, but
- Receive a failure from downstream Media API due to missing `UploadId` in validators.

Also, `FilesApiController.PostChunksComplete` **does not validate** `UploadId` at the MVC level:

```csharp
if (string.IsNullOrWhiteSpace(model.Filename) is true)
{
    return this.StatusCode((int)HttpStatusCode.UnprocessableEntity);
}

var fileId = await this.filesRepository.SaveChunksCompleteAsync(..., model.UploadId);
```

If `UploadId` is missing or invalid, the user just gets a generic error from deeper layers.

**Impact:**
- Potential breaking change for any other client not updated with the new `uploadId` semantics.
- Poor user-facing diagnostics when `UploadId` is invalid or missing.

**Recommendation:**
- Confirm that **all** callers of these endpoints are updated to send `uploadId`.
- Add explicit MVC-level validation in `FilesApiController` for `UploadId` and return a clear `400`/`422` with a meaningful message.
- Consider a short backward-compat window (e.g., allow missing `UploadId` and fall back to old behavior) if you have external clients.

---
## 4. HIGH – Type inconsistency: `UploadId` is `string` in some layers, `Guid` in others

**Examples:**

- **Frontend**: generates `uploadId` with `uuidv4()` – returns a **string**.
- **Seller.Web**: `FilesApiController.PostChunk` receives `string uploadId`, passes it to:

  ```csharp
  public async Task SaveChunkAsync(..., int? chunkNumber, string uploadId)
  {
      var requestModel = new FileRequestModelBase
      {
          UploadId = uploadId,
          ...
      };
  }
  ```

- **ApiExtensions**: `FileRequestModelBase.UploadId` is `string`, and `ApiClientService` sends it as a form string field.
- **Media.Api**: expects `Guid? UploadId` and attempts to use it as such.

This does work as long as:

- The frontend only ever sends values that are parseable `Guid`s.

But the contract is implicit and fragile: nothing prevents the FE from changing to another UUID format without hyphens or adding a prefix.

**Impact:**
- Hidden coupling between FE string format and BE Guid parsing.
- Harder to reason about and more error-prone over time.

**Recommendation:**
- Conceptually, `UploadId` looks like a server-side **identifier**, so modelling it as a `Guid` end-to-end is reasonable.
- At a minimum, document clearly that `uploadId` **must** be a valid Guid string.
- Better: for new code, expose uploadId from a BE-generated Guid (e.g., return initial `uploadId` from first call) rather than FE-generated.

---
## 5. MEDIUM – `Directory.GetFiles` error path vs intended `InvalidOperationException`

**Location:** `MediaService.CreateFileFromChunksAsync` & `UpdateFileFromChunksAsync`

```csharp
var directory = Path.Combine(
        MediaConstants.Paths.TempPath,
        model.OrganisationId?.ToString());

var searchPattern = $"{model.UploadId}_*";

string[] filePaths = Directory.GetFiles(directory, searchPattern)
    .OrderBy(...)
    .ToArray();

if (!filePaths.Any())
    throw new InvalidOperationException($"No chunks found for upload {model.UploadId}");
```

If `directory` does **not exist**, `Directory.GetFiles` throws `DirectoryNotFoundException` **before** your `if (!filePaths.Any())` check, so your nice `InvalidOperationException("No chunks found…")` is never used.

**Impact:**
- Error behavior is inconsistent and harder to diagnose.
- In edge cases (wrong org, uploadId, or premature `chunkssavecomplete`), you get a low-level exception instead of the intended friendly one.

**Recommendation:**
- Guard the directory:

  ```csharp
  if (!Directory.Exists(directory))
      throw new InvalidOperationException($"No chunks found for upload {model.UploadId}");
  ```

  before calling `Directory.GetFiles`.

---
## 6. LOW–MEDIUM – Minor issues in the Media service implementation

1. **Redundant `stream.Close()` inside `using` block**

   ```csharp
   await using var stream = new FileStream(newPath, FileMode.Open, FileAccess.Read, FileShare.Read);
   ...
   stream.Close();
   File.Delete(newPath);
   ```

   - The `await using` already ensures `Dispose()` / `Close()` is called.
   - Explicit `Close()` is redundant; in a worst case, if someone refactors and reorders, it could cause unexpected behavior.

   **Recommendation:** Remove the explicit `stream.Close()`.

2. **Chunk sorting assumes a specific filename format**

   ```csharp
   var name = Path.GetFileName(p);
   var parts = name.Split('_');
   return int.Parse(parts[1]);
   ```

   - This assumes exactly one `_` separator and that `parts[1]` is always a valid int.
   - Given current naming (`{Guid}_{chunkNumber}`), this is OK, but any future change to naming will break ordering.

   **Recommendation:** At least add basic validation/guard or a comment describing the required format.

---
## 7. LOW – Frontend UX / robustness issues

1. **Upload progress may not reach 100% cleanly**

   - For the last chunk, you call `uploadInChunksComplete()` directly when `chunkCounter + 1 === chunkCount` and do not explicitly set `fileUploadProgress` to 100.
   - In practice you clear the UI after completion, but from a UX standpoint it’s nicer and clearer to show 100% before resetting.

2. **State/reset logic is a bit fragile**

   - There’s a lot of tightly coupled state (`chunkCounter`, `chunkCount`, `beginingOfTheChunk`, `endOfTheChunk`, `uploadId`) being reset in several places.
   - This made it easy to introduce the critical `uploadId` bug for multi-file uploads.

   **Recommendation:** Consider encapsulating per-file upload state in an object (or hook) instead of many independent `useState` calls.

---
## 8. LOW – Style / consistency nits

- `UploadId` nullability is inconsistent even where validators require it.
- The existing typo `ChunkSumber` remains; not introduced here but worth fixing eventually for clarity.
- Some `using` patterns are mixed (`using` vs `await using`), which is mostly stylistic but can be standardized.

---
## Summary

**Most important to fix before merging:**
1. Ensure a new, valid `uploadId` is generated for **each file** and propagated consistently (remark 1).
2. Make `UploadId` handling consistent and safe across layers – avoid nullable `Guid?` where it’s logically required, and don’t dereference without checks (remark 2).
3. Double‑check and document the breaking-change aspect for existing clients and validate `UploadId` earlier with clear error messages (remark 3).

If you’d like, I can propose a concrete refactor for `MediaCloud.js` that fixes the per-file `uploadId` handling and simplifies the chunk state machine.

dawiddworak88 (changes requested):
# Code Review: File Upload Chunking PR

I've analyzed the pull request that implements a file upload chunking system with a unique `uploadId` to prevent file collision issues. Here's my comprehensive code review, ranked by importance:

---

## 🔴 **CRITICAL ISSUES**

### 1. **Race Condition & File Collision Risk in Chunk Cleanup** (MediaService.cs)
**Location:** `CreateFileFromChunksAsync` and `UpdateFileFromChunksAsync` methods

**Problem:** The PR removes the immediate deletion of chunk files after merging them, but then deletes them all at once after the final file is created. However, there's a critical issue:

```csharp
foreach (var filePath in filePaths)
{
    MergeChunks(newPath, filePath);
    // Chunk deletion removed from here - GOOD
}

// ... file processing ...

File.Delete(newPath); // Only deletes the merged file, NOT the chunks!
```

**Impact:** Chunk files are never cleaned up! The `MergeChunks` method now deletes the source chunk after merging, but if an exception occurs during file processing (between merging and final cleanup), chunks will remain on disk indefinitely, causing disk space issues.

**Recommendation:** Add proper cleanup in a try-finally block or ensure chunks are deleted after successful merge.

---

### 2. **Incorrect Polish Translation** (MediaResources.pl.resx)
**Location:** Line 145

```xml
<data name="ChunkDirectoryNotFound" xml:space="preserve">
    <value>Dodaj media</value>  <!-- This means "Add media" -->
</data>
```

**Problem:** The translation is completely wrong. It should say "Katalog chunków nie znaleziony" or similar, not "Add media" (which is already used for `NewText`).

**Impact:** Users will see confusing error messages in Polish.

---

### 3. **Missing Validation for UploadId in Frontend** (MediaCloud.js)
**Location:** Lines 141-145

**Problem:** The frontend generates a `uploadId` using `uuidv4()`, but there's no validation to ensure it was successfully generated before sending it to the backend.

```javascript
const newUploadId = uuidv4();
setUploadId(newUploadId);
```

**Impact:** If `uuidv4()` fails (unlikely but possible), the upload will proceed with an undefined/null uploadId, causing backend validation failures.

**Recommendation:** Add validation:
```javascript
const newUploadId = uuidv4();
if (!newUploadId) {
    toast.error(props.generalErrorMessage);
    return;
}
setUploadId(newUploadId);
```

---

## 🟡 **HIGH PRIORITY ISSUES**

### 4. **Inconsistent Type for UploadId** (FileRequestModelBase.cs)
**Location:** Line 5

**Problem:** `UploadId` is defined as `string` in `FileRequestModelBase`, but as `Guid` everywhere else (service models, validators, request models).

```csharp
public class FileRequestModelBase
{
    public string Id { get; set; }
    public string UploadId { get; set; }  // Should be Guid
    // ...
}
```

**Impact:** This creates type inconsistency and potential parsing issues. The API client service checks `!string.IsNullOrWhiteSpace(request.Data.UploadId)` which works for strings, but the backend expects a Guid.

**Recommendation:** Change to `Guid UploadId { get; set; }` and update the validation in ApiClientService.cs to check for `Guid.Empty`.

---

### 5. **Memory Inefficiency in File Streaming** (MediaService.cs)
**Location:** Lines 502-504 (new code)

**Problem:** The code now uses `FileStream` instead of `MemoryStream`, which is good, but it doesn't dispose of the stream properly in the old code path:

```csharp
await using var stream = new FileStream(newPath, FileMode.Open, FileAccess.Read, FileShare.Read);
var formFile = new FormFile(stream, 0, stream.Length, model.Filename, model.Filename);
```

**Issue:** The `FormFile` holds a reference to the stream, but the stream is disposed immediately after the `using` block ends, potentially before the file is fully processed by `CreateFileAsync`.

**Recommendation:** Review the lifecycle of the FormFile and ensure the stream remains open during the entire file creation process.

---

### 6. **Potential Directory Traversal Vulnerability**
**Location:** `CreateFileChunkAsync` method (lines 449-453)

**Problem:** The `uploadId` comes from user input and is used directly in the file path:

```csharp
var fileName = $"{model.UploadId}_{model.ChunkSumber}";
var path = Path.Combine(directory, fileName);
```

**Impact:** While `uploadId` is validated as a Guid (which limits the attack surface), if validation is bypassed or the type is changed to string, this could allow directory traversal attacks.

**Recommendation:** Add explicit path validation:
```csharp
var fileName = $"{model.UploadId}_{model.ChunkSumber}";
if (fileName.Contains("..") || fileName.Contains(Path.DirectorySeparatorChar))
{
    throw new CustomException("Invalid file name", (int)HttpStatusCode.BadRequest);
}
```

---

## 🟢 **MEDIUM PRIORITY ISSUES**

### 7. **Improved Error Handling but Missing Cleanup** (MediaService.cs)
**Location:** Lines 475-477, 509-511

**Problem:** The new code properly throws exceptions when chunks are not found:

```csharp
if (!Directory.Exists(directory))
    throw new CustomException(_mediaResources.GetString("ChunkDirectoryNotFound"), (int)HttpStatusCode.UnprocessableEntity);

if (!filePaths.Any())
    throw new CustomException(_mediaResources.GetString("NoChunksFound"), (int)HttpStatusCode.UnprocessableEntity);
```

**Issue:** If an exception is thrown, any partially uploaded chunks remain on disk. There's no cleanup mechanism.

**Recommendation:** Implement a cleanup mechanism or document that chunks should be cleaned up by a background job.

---

### 8. **Chunk Counter Logic Improvement** (MediaCloud.js)
**Location:** Lines 149-151

**Problem:** The chunk counter logic has been improved, but the condition check is still slightly confusing:

```javascript
if (chunkCounter < chunkCount) {  // Changed from <=
    const chunk = fileToUploadInChunks.slice(beginingOfTheChunk, endOfTheChunk);
    uploadChunk(chunk);
}
```

**Issue:** The logic is correct now, but the state management is complex with multiple useEffect dependencies.

**Recommendation:** Consider using a more explicit state machine or reducer pattern for chunk upload state management.

---

### 9. **Missing Null Check** (FilesApiController.cs)
**Location:** Lines 135-138

**Problem:** The controller validates parameters but doesn't check if the file chunk is empty:

```csharp
if (chunk is not null && chunkNumber.HasValue && 
    string.IsNullOrWhiteSpace(filename) is false &&
    string.IsNullOrWhiteSpace(uploadId) is false)
```

**Issue:** A chunk could be non-null but have zero length, which should be rejected.

**Recommendation:** Add `chunk.Length > 0` to the validation.

---

### 10. **Inconsistent Async Pattern** (MediaService.cs)
**Location:** `CreateFileChunkAsync` method (line 453)

**Problem:** The method uses `await using` for the FileStream, which is good, but the old code used `using` without await:

```csharp
await using var fs = File.Create(path);
await model.File.CopyToAsync(fs);
```

**Issue:** While this is correct, it's inconsistent with the old pattern. The improvement is good, but should be documented.

---

## 🔵 **LOW PRIORITY / STYLE ISSUES**

### 11. **Typo in Variable Name** (Throughout)
**Location:** Multiple files

**Problem:** `ChunkSumber` should be `ChunkNumber` (typo in "Number").

**Impact:** Inconsistent naming, but doesn't affect functionality.

---

### 12. **Improved MergeChunks Method** (MediaService.cs)
**Location:** Lines 556-565

**Positive Change:** The new implementation is much cleaner:

```csharp
using (var fs1 = new FileStream(targetPath, FileMode.Append, FileAccess.Write, FileShare.None))
using (var fs2 = new FileStream(sourcePath, FileMode.Open, FileAccess.Read, FileShare.Read))
{
    fs2.CopyTo(fs1);
}
File.Delete(sourcePath);
```

This is a significant improvement over the old manual byte array copying.

---

### 13. **Frontend State Management Complexity** (MediaCloud.js)
**Location:** Lines 141-280

**Issue:** The chunk upload logic has many interdependent state variables (`uploadId`, `chunkCounter`, `chunkCount`, `fileToUploadInChunks`, etc.), making it difficult to reason about the state flow.

**Recommendation:** Consider refactoring to use a reducer pattern or extracting the chunk upload logic into a custom hook.

---

## ✅ **POSITIVE CHANGES**

1. **UploadId Implementation:** Solves the file collision problem effectively
2. **Better File Streaming:** Using `FileStream` instead of loading entire files into memory
3. **Improved Error Messages:** New localized error messages for chunk operations
4. **Better Chunk Ordering:** Using `uploadId_chunkNumber` pattern instead of filename-based ordering
5. **Validation Improvements:** Added validators for `UploadId` in all service models
6. **Cleaner MergeChunks:** Much more efficient implementation

The PR successfully addresses the file collision problem with the `uploadId` approach, but has several critical issues that need to be resolved, particularly around chunk cleanup and type consistency.

dawiddworak88 (approved):

SebaKas21:
[celebrate]     Sebastian Kasperek reacted to your message:
________________________________

PR #518, opened 2026-01-27
Correct calculate stock and outlet quantity

PiotrZaranek:


dawiddworak88 (changes requested):
## Code Review: Available Quantities Fix for Inventory/Outlet Stock

I've thoroughly analyzed the diff and the related source files. Here's my comprehensive code review:

---

### **Summary of Changes**

The diff modifies three files to fix how available quantities are updated when a basket checkout occurs:

1. **BasketService.cs** - Changes how `BookedQuantity` is calculated
2. **InventoryService.cs** - Changes `UpdateInventoryQuantity` method logic
3. **OutletService.cs** - Changes `UpdateOutletQuantity` method logic

---

### **🔴 CRITICAL ISSUES**

#### 1. **BasketService.cs - Sign Change is INCORRECT**

**Current code (before diff):**
```csharp
BookedQuantity = (int)-group.Sum(x => x.StockQuantity)
```

**Proposed change:**
```csharp
BookedQuantity = (int)group.Sum(x => x.StockQuantity)
```

**Problem:** The original code negates the quantity (making it negative) because the old `UpdateInventoryQuantity` method was **adding** the booked quantity to the inventory. The logic was:
- `inventory.Quantity + bookedQuantity` where `bookedQuantity` was negative → effectively subtracting

The proposed change removes the negation, which is correct **only if** the receiving method now subtracts instead of adds. Let me verify this is consistent...

✅ **This is consistent** with the changes in `InventoryService.cs` and `OutletService.cs` which now subtract: `inventory.AvailableQuantity - bookedQuantity`

---

#### 2. **InventoryService.cs - Logic Issues**

**Proposed change:**
```csharp
public async Task UpdateInventoryQuantity(Guid? productId, double bookedQuantity)
{
    if (bookedQuantity > 0) 
    {
        var inventory = _context.Inventory.FirstOrDefault(x => x.ProductId == productId.Value && x.IsActive);

        if (inventory is not null)
        {
            var productQuantity = inventory.AvailableQuantity - bookedQuantity;

            if (productQuantity < 0)
            {
                productQuantity = 0;
            }

            inventory.AvailableQuantity = productQuantity;
            inventory.LastModifiedDate = DateTime.UtcNow;

            await _context.SaveChangesAsync();
        }
    }
}
```

**Issues Found:**

| Issue | Severity | Description |
|-------|----------|-------------|
| **Missing `inventory.Quantity` update** | 🔴 High | The original code updated BOTH `inventory.Quantity` and `inventory.AvailableQuantity`. The new code only updates `AvailableQuantity`. This may cause data inconsistency between `Quantity` and `AvailableQuantity` fields. |
| **Potential null reference** | 🟡 Medium | `productId.Value` is accessed without null check on `productId`. If `productId` is null, this will throw `InvalidOperationException`. |
| **Missing async on FirstOrDefault** | 🟡 Medium | `FirstOrDefault` should be `FirstOrDefaultAsync` for consistency with async pattern. |
| **Multiple inventory records not handled** | 🟡 Medium | If a product exists in multiple warehouses, only the first inventory record is updated. The original code had the same issue, but this should be addressed. |

---

#### 3. **OutletService.cs - Same Logic Issues**

The same issues apply to `OutletService.cs`:

| Issue | Severity | Description |
|-------|----------|-------------|
| **Missing `outlet.Quantity` update** | 🔴 High | Same as above - only `AvailableQuantity` is updated, not `Quantity`. |
| **Potential null reference** | 🟡 Medium | Same `productId.Value` issue. |
| **Missing async on FirstOrDefault** | 🟡 Medium | Should use `FirstOrDefaultAsync`. |

---

### **🟡 DESIGN CONCERNS**

#### 1. **Quantity vs AvailableQuantity Semantics**

Looking at the codebase, there appear to be two fields:
- `Quantity` - Total physical quantity in stock
- `AvailableQuantity` - Quantity available for sale (may differ due to reservations)

**Question:** Should checkout reduce:
- Only `AvailableQuantity` (items are reserved but not yet shipped)?
- Both `Quantity` and `AvailableQuantity` (items are immediately considered sold)?

The original code updated both. The new code only updates `AvailableQuantity`. **This needs clarification on business requirements.**

#### 2. **Race Condition Risk**

The code reads inventory, calculates new quantity, then saves. In high-concurrency scenarios, two simultaneous checkouts could read the same `AvailableQuantity`, both subtract their amounts, and one update would be lost.

**Recommendation:** Consider using optimistic concurrency or database-level atomic operations.

---

### **🟢 POSITIVE ASPECTS**

1. **Guard clause added** - `if (bookedQuantity > 0)` prevents unnecessary database operations when quantity is zero
2. **Cleaner logic** - Subtracting a positive number is more intuitive than adding a negative number
3. **Consistent pattern** - Both `InventoryService` and `OutletService` follow the same pattern

---

### **RECOMMENDED FIXES**

```csharp
// InventoryService.cs - Recommended implementation
public async Task UpdateInventoryQuantity(Guid? productId, double bookedQuantity)
{
    if (productId.HasValue && bookedQuantity > 0) 
    {
        var inventory = await _context.Inventory
            .FirstOrDefaultAsync(x => x.ProductId == productId.Value && x.IsActive);

        if (inventory is not null)
        {
            var newAvailableQuantity = inventory.AvailableQuantity - bookedQuantity;
            var newQuantity = inventory.Quantity - bookedQuantity;

            inventory.AvailableQuantity = Math.Max(0, newAvailableQuantity);
            inventory.Quantity = Math.Max(0, newQuantity);  // Keep this if business requires
            inventory.LastModifiedDate = DateTime.UtcNow;

            await _context.SaveChangesAsync();
        }
    }
}
```

---

### **QUESTIONS FOR CLARIFICATION**

1. **Should `Quantity` also be decremented?** The original code updated both fields. Is this intentional to only update `AvailableQuantity` now?

2. **What about multiple warehouse inventory?** If a product exists in multiple warehouses, should the quantity be deducted from a specific warehouse or distributed?

3. **Is there a separate process to reconcile `Quantity` and `AvailableQuantity`?** For example, when orders are shipped?

---

### **VERDICT**

The fix addresses the core issue (quantities not being properly decremented on checkout), but has some gaps that should be addressed before merging:

1. ⚠️ Decide whether to keep updating `Quantity` field
2. ⚠️ Add null check for `productId`
3. ⚠️ Use `FirstOrDefaultAsync` instead of `FirstOrDefault`

Would you like me to provide the corrected implementation once you clarify the business requirements?

dawiddworak88 (approved):

PR #522, opened 2026-02-04
Correct calculating qunatity in order form and method to get outelts by productids

PiotrZaranek:


be/src/Project/Services/Inventory/Inventory.Api/Services/InventoryItems/InventoryService.cs
```diff
@@ -499,23 +499,29 @@ public PagedResults<IEnumerable<InventorySumServiceModel>> GetAvailableProductsI
...
-                {
-                    productQuantity = 0;
-                }
+            if (inventories.Any() is false) return;
 
-                inventory.AvailableQuantity = productQuantity;
-                inventory.LastModifiedDate = DateTime.UtcNow;
+            var totalAvailableQuantity = inventories.Sum(x => x.AvailableQuantity);
+            if (bookedQuantity > totalAvailableQuantity)
+                throw new ConflictException("Insufficient stock");
```
SebaKas21: **suggestion** Missing translation

be/src/Project/Services/Inventory/Inventory.Api/Services/InventoryItems/InventoryService.cs
```diff
@@ -499,23 +499,29 @@ public PagedResults<IEnumerable<InventorySumServiceModel>> GetAvailableProductsI
...
-                var productQuantity = inventory.AvailableQuantity - bookedQuantity;
+            var inventories = await _context.Inventory
+                .Where(x => x.ProductId == productId && x.IsActive)
+                .ToListAsync();
 
-                if (productQuantity < 0)
-                {
-                    productQuantity = 0;
-                }
+            if (inventories.Any() is false) return;
```
SebaKas21: **suggestion** inventories could be empty add OrEmptyIfNull, let's try !inventories.OrEmptyIfNull().Any()

be/src/Project/Services/Inventory/Inventory.Api/Services/OutletItems/OutletService.cs
```diff
@@ -550,42 +550,63 @@ public async Task UpdateOutletQuantity(Guid? productId, double bookedQuantity)
...
+                x.ProductId,
+                ProductName = x.Product.Name,
+                ProductEan = x.Product.Ean,
+                ProductSku = x.Product.Sku,
+                Translation = x.Translations.FirstOrDefault(t => t.OutletItemId == x.Id && t.Language == model.Language)
+                  ?? x.Translations.FirstOrDefault(t => t.OutletItemId == x.Id),
+                x.Quantity,
+                x.AvailableQuantity
+            })
+            .AsEnumerable()
```
SebaKas21: **question** AsEnumerable is required?

SebaKas21 (changes requested):

dawiddworak88 (changes requested):
## Code Review: Multi-Warehouse Inventory & Outlet Fix

I identified the following remarks, **ranked by priority** (Critical → High → Medium → Low):

---

### 🔴 CRITICAL PRIORITY

#### 1. **Removal of Stock/Outlet Validation Before Checkout (Security/Data Integrity Risk)**
**Files:** `BasketCheckoutApiController.cs`, `BasketService.cs`, `IBasketService.cs`

The `ValidateStockOutletQuantitiesAsync` method has been completely removed, along with its call in the checkout flow:
```csharp
// REMOVED:
await _basketService.ValidateStockOutletQuantitiesAsync(model.BasketId, token, language);
```

**Impact:** This is a **critical regression**. Without this validation:
- Users can place orders for quantities exceeding available stock
- Race conditions between concurrent orders are now unhandled at the checkout level
- The only protection is now the `ConflictException` thrown during inventory deduction, which happens **after** order processing begins

**Recommendation:** Either:
1. Restore the validation (preferred for UX - fail fast with clear error)
2. Ensure the `ConflictException` in `UpdateInventoryQuantity`/`UpdateOutletQuantity` is properly caught and handled in the checkout flow with appropriate rollback

---

### 🟠 HIGH PRIORITY

#### 2. **No Ordering Strategy for Multi-Warehouse Deduction**
**Files:** `InventoryService.cs`, `OutletService.cs`

The new allocation logic iterates through inventories without any defined order:
```csharp
var inventories = await _context.Inventory
    .Where(x => x.ProductId == productId && x.IsActive)
    .ToListAsync();
// No OrderBy - order is undefined!

foreach (var item in inventories)
{
    var toDeduct = Math.Min(item.AvailableQuantity, remainingToAllocate);
    ...
}
```

**Impact:** 
- Deduction order is non-deterministic (depends on DB order)
- Could lead to inconsistent warehouse depletion patterns
- May cause issues with warehouse prioritization (e.g., nearest warehouse first)

**Recommendation:** Add explicit ordering, e.g.:
```csharp
.OrderByDesc(x => x.CreatedDate) 
```

---

#### 3. **Potential Null Reference in Buyer ProductsApiController**
**File:** `Buyer.Web/Areas/Products/ApiControllers/ProductsApiController.cs`

```csharp
StockQuantity = inventories.Where(x => x.ProductId == product.Id).Sum(x => x.AvailableQuantity) ?? 0,
```

**Issue:** `Sum()` on `double` returns `double`, not `double?`. The `?? 0` is unnecessary and suggests the code may have been copied incorrectly. If `AvailableQuantity` is nullable (`double?`), the `Sum()` would return `double?`, but this needs verification.

**Note:** The Seller version doesn't have `?? 0`, creating inconsistency between the two controllers.

---

### 🟡 MEDIUM PRIORITY

#### 4. **Frontend Validation Logic Change May Allow Invalid States**
**Files:** `NewOrderForm.js`, `OrderForm.js`

The validation logic was changed:
```javascript
// BEFORE (onBlur):
if (isNaN(numericValue) || numericValue < 1) {
    numericValue = 1;
}
if (productFromOutlet && numericValue > maxOutlet) {
    numericValue = maxOutlet;
}

// AFTER (onChange + onBlur):
onChange: if (productFromOutlet && e.target.value > maxOutlet) {
    e.target.value = maxOutlet;
}
onBlur: if (isNaN(numericValue) || numericValue < 0) {  // Changed from < 1 to < 0
    numericValue = 1;
}
```

**Issues:**
1. `numericValue < 0` allows `0` as valid, but then sets to `1` - inconsistent
2. The `onChange` handler mutates `e.target.value` directly (anti-pattern in React)
3. Max outlet validation moved to `onChange` only - won't catch programmatic changes
4. String comparison in `e.target.value > maxOutlet` may not work correctly (e.g., "9" > 10 is true as string comparison)

**Recommendation:** Use proper numeric comparison and keep validation consistent.

---

#### 5. **GetOutletsByProductsIds Query Change - Performance Consideration**
**File:** `OutletService.cs`

The query was refactored from a LINQ join to method syntax with `.Include()`:
```csharp
var outletItems = _context.Outlet
    .Include(x => x.Translations)
    .Include(x => x.Product)
    .Include(x => x.Warehouse)
    .AsSingleQuery()
    ...
```

**Observations:**
- `Warehouse` is included but never used in the projection
- The new query uses `AsEnumerable()` which materializes results before the final projection
- Translation fallback logic (`?? x.Translations.FirstOrDefault(...)`) is good for i18n

**Recommendation:** Remove unused `.Include(x => x.Warehouse)` to reduce query overhead.

---

### 🟢 LOW PRIORITY

#### 6. **Inconsistent Error Messages**
**Files:** `InventoryService.cs`, `OutletService.cs`

```csharp
// InventoryService:
throw new ConflictException("Insufficient stock");

// OutletService:
throw new ConflictException("Insufficient outlet");
```

**Issue:** "Insufficient outlet" is grammatically awkward. Should be "Insufficient outlet stock" or similar.

---

#### 7. **Unused Import Removal**
**File:** `BasketService.cs`

```csharp
// REMOVED:
using System.Text.Json;
```

This is fine - good cleanup of unused imports.

---

#### 8. **Minor: Inconsistent Line Break in Query**
**File:** `OutletService.cs`

```csharp
var outlets = await _context.Outlet.  // Dot on wrong line
    Where(x => x.ProductId == productId.Value && x.IsActive)
```

Should be:
```csharp
var outlets = await _context.Outlet
    .Where(x => x.ProductId == productId.Value && x.IsActive)
```

---

## Summary Table

| Priority | Issue | Files Affected |
|----------|-------|----------------|
| 🔴 Critical | Validation removed from checkout | BasketCheckoutApiController, BasketService |
| 🟠 High | No ordering for multi-warehouse deduction | InventoryService, OutletService |
| 🟠 High | Potential null reference / inconsistency | ProductsApiController (Buyer vs Seller) |
| 🟡 Medium | Frontend validation logic issues | NewOrderForm.js, OrderForm.js |
| 🟡 Medium | Unused Include in query | OutletService |
| 🟢 Low | Inconsistent error messages | InventoryService, OutletService |
| 🟢 Low | Minor code style issues | OutletService |

---

dawiddworak88 (changes requested):
# Code Review: Multi-Warehouse Inventory Deduction for Orders

I've thoroughly reviewed the diff and the surrounding source files. Below are my findings ranked by severity.

---

## 🔴 CRITICAL (Severity 1) — Race Condition / No Concurrency Control on Inventory Deduction

**Files:** `InventoryService.cs` → `UpdateInventoryQuantity`, `OutletService.cs` → `UpdateOutletQuantity`

The new multi-warehouse deduction logic loads all inventory rows into memory (`.ToListAsync()`), modifies them in a loop, then calls `SaveChangesAsync()` — all without any concurrency protection (no transaction isolation level, no optimistic concurrency tokens, no `SELECT ... FOR UPDATE` equivalent).

**Problem:** If two orders for the same product are placed simultaneously, both will read the same `AvailableQuantity` values, both will deduct, and the database will end up with **oversold inventory** (negative effective stock). This is the most dangerous issue in the entire changeset.

**Recommendation:**
- Use **optimistic concurrency** (EF Core `[ConcurrencyCheck]` or `[Timestamp]`/`rowversion` on the inventory entity) so that a concurrent modification throws `DbUpdateConcurrencyException`, which can be retried.
- Alternatively, wrap the read + update in an explicit **serializable transaction** or use a raw SQL `UPDATE ... SET AvailableQuantity = AvailableQuantity - @deduct WHERE AvailableQuantity >= @deduct` to make it atomic.

---

## 🔴 CRITICAL (Severity 2) — `.ToList()` Materializes Entire Table Subset in `GetOutletsByProductsIds`

**File:** `OutletService.cs` → `GetOutletsByProductsIds`

```csharp
.ToList()
.Select(x => new OutletSumServiceModel { ... })
```

The old code used a LINQ-to-SQL query (translated to SQL). The new code calls `.ToList()` which **materializes all matching outlet rows + their translations + products into memory** before projecting. If `model.Ids` contains many product IDs, this could load thousands of rows with all their navigation properties.

**Recommendation:** Keep the projection server-side. Use `.Select(...)` before `.ToList()` so EF Core translates the projection to SQL. If the `?.` null-conditional operators or `FirstOrDefault` with language fallback can't be translated, consider using a raw SQL query or a two-pass approach (IDs first, then details).

---

## 🟠 HIGH (Severity 3) — Behavioral Change: Silent Clamp to Zero → Hard Exception

**Files:** `InventoryService.cs`, `OutletService.cs`

**Old behavior:** If `bookedQuantity > availableQuantity`, the system would silently clamp to 0 (allow the order, just zero out stock).  
**New behavior:** Throws `ConflictException` and **rejects the order entirely**.

This is a significant behavioral change. While the new behavior is arguably more correct, it means:
- Orders that previously succeeded will now **fail** if stock is insufficient.
- There's no partial fulfillment — if total across all warehouses is 9 and order is 10, the entire order is rejected.
- The error message is generic and doesn't tell the user *how much* is available.

**Recommendation:** Confirm this is the desired business behavior. Consider including the available quantity in the error message. Also consider whether partial fulfillment should be supported.

---

## 🟠 HIGH (Severity 4) — Allocation Strategy is Arbitrary (Newest Warehouse First)

**Files:** `InventoryService.cs`, `OutletService.cs`

```csharp
.OrderByDescending(x => x.CreatedDate)
```

Inventory is deducted from the **most recently created** warehouse record first. This is an implicit business rule with no documentation or configurability.

**Recommendation:**
- Document why "newest first" was chosen. Common strategies are FIFO (oldest first), closest warehouse, or warehouse with most stock.
- Consider making this configurable or at least adding a code comment explaining the rationale.

---

## 🟡 MEDIUM (Severity 5) — Polish Localization Typo

**File:** `InventoryResources.pl.resx`

```
"Próbowano złożyć zamówienie na więszką ilość niż jest dostępna."
```

`więszką` should be `**większą**` (missing `k` before `sz` → should be `ksz`).

---

## 🟡 MEDIUM (Severity 6) — Redundant `.OrEmptyIfNull()` on Non-Nullable List

**Files:** `InventoryService.cs`, `OutletService.cs`

```csharp
var inventories = await _context.Inventory...ToListAsync();
if (inventories.OrEmptyIfNull().Any() is false) return;
```

`ToListAsync()` never returns `null` — it always returns an empty list at minimum. The `.OrEmptyIfNull()` call is unnecessary and misleading (suggests the value could be null when it can't).

**Recommendation:** Simplify to `if (!inventories.Any()) return;`

---

## 🟡 MEDIUM (Severity 7) — `double` Used for Quantities (Pre-existing, but Amplified)

**Files:** `InventoryService.cs`, `OutletService.cs`, domain models

Quantities are `double`, which has floating-point precision issues. The new code does `Math.Min`, subtraction, and `Sum` across multiple warehouse rows, amplifying the risk of floating-point drift (e.g., `10.0 - 3.0000000000000004`). The comparison `bookedQuantity > totalAvailableQuantity` could produce false positives/negatives.

**Recommendation:** This is pre-existing tech debt, but the multi-warehouse summing makes it worse. Consider migrating to `decimal` for monetary/quantity calculations in a future iteration.

---

## 🟡 MEDIUM (Severity 8) — Nullable-to-Non-Nullable Breaking Change on DTOs

**Files:** `InventorySumResponseModel.cs`, `InventorySum.cs`, `OutletSum.cs`

```csharp
- public double? Quantity { get; set; }
- public double? AvailableQuantity { get; set; }
+ public double Quantity { get; set; }
+ public double AvailableQuantity { get; set; }
```

Changing from `double?` to `double` is a **breaking change** for any API consumers that expect nullable JSON fields. If external systems or the frontend check for `null`, they'll now get `0` instead, which has different semantics ("no data" vs "zero stock").

**Recommendation:** Verify no API consumers rely on the nullable semantics. If this is a public API, consider versioning.

---

## 🟢 LOW (Severity 9) — Frontend: `onChange` Clamping Only for Outlet, Not Stock

**Files:** `NewOrderForm.js`, `OrderForm.js`

The new `onChange` handler clamps quantity to `maxOutlet` when `productFromOutlet` is true, but there's **no clamping for stock quantity**. A user could type a stock quantity exceeding available stock — the validation only happens server-side on order placement.

**Recommendation:** Consider adding client-side clamping for stock quantities too for a better UX, or document that server-side validation is the intentional guard.

---

## 🟢 LOW (Severity 10) — Seller `OrderForm.js` Not Updated with Same Input Validation

**File:** `OrderForm.js` (Seller.Portal)

The Seller portal's `OrderForm.js` still has the **old** `onChange` handler (`setQuantity(e.target.value)`) in the current codebase, but the diff shows it being updated. However, the Seller form's `maxOutlet` calculation is simpler (`product ? product.outletQuantity : 0`) and doesn't account for items already in the basket (unlike the Buyer form which uses `QuantityCalculatorService.calculateMaxQuantity`). This inconsistency could allow the seller to over-order outlet items.

**Recommendation:** Align the `maxOutlet` calculation logic between Buyer and Seller forms.

---

## 🟢 LOW (Severity 11) — Minor Code Quality: Double Semicolon Fix

**File:** `OutletService.cs`

```csharp
- });;
+ });
```

Good catch — this was a harmless syntax issue but good to clean up.

---

## 🟢 LOW (Severity 12) — Whitespace Fix

**File:** `OutletService.cs`

```csharp
- Title= outletItems...
+ Title = outletItems...
```

Cosmetic fix, good for consistency.

---

## Summary Table

| # | Severity | Issue | File(s) |
|---|----------|-------|---------|
| 1 | 🔴 CRITICAL | Race condition — no concurrency control on inventory deduction | `InventoryService.cs`, `OutletService.cs` |
| 2 | 🔴 CRITICAL | `.ToList()` materializes full dataset into memory | `OutletService.cs` (`GetOutletsByProductsIds`) |
| 3 | 🟠 HIGH | Silent clamp → hard exception (behavioral change) | `InventoryService.cs`, `OutletService.cs` |
| 4 | 🟠 HIGH | Arbitrary allocation strategy (newest first) undocumented | `InventoryService.cs`, `OutletService.cs` |
| 5 | 🟡 MEDIUM | Polish typo: "więszką" → "większą" | `InventoryResources.pl.resx` |
| 6 | 🟡 MEDIUM | Redundant `.OrEmptyIfNull()` on non-nullable list | `InventoryService.cs`, `OutletService.cs` |
| 7 | 🟡 MEDIUM | `double` precision risk amplified by multi-warehouse summing | Multiple files |
| 8 | 🟡 MEDIUM | Nullable→non-nullable DTO breaking change | `InventorySumResponseModel.cs`, `InventorySum.cs`, `OutletSum.cs` |
| 9 | 🟢 LOW | No client-side stock quantity clamping | `NewOrderForm.js`, `OrderForm.js` |
| 10 | 🟢 LOW | Inconsistent `maxOutlet` calculation between Buyer/Seller | `OrderForm.js` vs `NewOrderForm.js` |
| 11 | 🟢 LOW | Double semicolon fix (good) | `OutletService.cs` |
| 12 | 🟢 LOW | Whitespace fix (good) | `OutletService.cs` |

dawiddworak88 (changes requested):
## Final recommendations for the PR review

### Corrections to the supplied review

### Must fix before merge

1. **Wrap each deduction loop in a transaction**

   `InventoryService.UpdateInventoryQuantity` and `OutletService.UpdateOutletQuantity` can currently partially deduct rows, then throw a conflict when a later row fails. Because these methods are likely called from ordering/integration-event flows, partial deductions followed by retry can cause double deduction or permanently inconsistent stock.

   Recommendation:

   - Begin a DB transaction before the allocation loop.
   - Commit only after all rows are updated successfully.
   - Roll back on `ConflictException` or any unexpected exception.

2. **Avoid tracked stale entities when using raw SQL**

   The methods load stock rows via tracked `ToListAsync()`, then mutate the same rows using raw SQL. The DbContext change tracker keeps stale `AvailableQuantity` values.

   Recommendation:

   - Add `.AsNoTracking()` to the allocation read queries, or
   - avoid raw SQL and use an EF-supported update pattern consistently.

3. **Fix/decide the concurrency behavior after CAS failure**

   The optimistic guard `AND AvailableQuantity >= {toDeduct}` prevents overselling a single row, but `toDeduct` is computed from a stale snapshot. If another transaction changes a row between read and update, the method immediately throws even if other rows still have enough quantity.

   Acceptable options:

   - Treat any CAS failure as a conflict, but ensure the transaction rolls back previous deductions; or
   - implement retry/spill-over allocation: re-read remaining live rows and continue while total availability still satisfies the booking.

   The first option is simpler and safe, but may produce more conflicts under load. The second option is more user-friendly but more complex.

4. **Decide FIFO vs LIFO allocation explicitly**

   The new logic consumes newest rows first via `OrderByDescending(x => x.CreatedDate)`. For inventory, FIFO is usually expected unless there is a business reason for LIFO.

   Recommendation:

   - Confirm with product/accounting stakeholders.
   - If not intentional, change to `OrderBy(x => x.CreatedDate)`.
   - Add a short comment documenting the chosen policy.

5. **Handle “no stock rows found” deliberately**

   Both update methods silently return when no active rows exist. For a booked product, that probably means inventory data is missing or stale.

   Recommendation:

   - Throw the same quantity conflict, or
   - at minimum log a warning/error with `productId` and `bookedQuantity`.

   Silent success is risky because it allows an order flow to proceed without deducting anything.

### Should fix before merge, unless explicitly accepted

6. **Simplify and constrain `GetOutletsByProductsIds`**

   The new shape — one outlet row rather than one row per translation — is directionally correct. But the implementation should be simplified:

   - Use a single query instead of fetching `outletIds` first.
   - Keep filtering in SQL.
   - Avoid repeated `FirstOrDefault` calls over `Translations`.
   - Remove redundant `t.OutletItemId == x.Id` inside the navigation collection.

   This is less critical than the allocator bugs, but worth fixing because this endpoint feeds basket/product quantity validation.

7. **Confirm DTO nullability change**

   Changing Buyer `Quantity` / `AvailableQuantity` from `double?` to `double` is a contract change. It is probably safe if all producers are controlled and always emit numbers, but it should be explicitly confirmed, especially for rolling deployments or external/mobile consumers.

8. **Add allocator tests**

   Given the concurrency-sensitive behavior, tests should be required for:

   - Allocation split across multiple rows.
   - Insufficient total quantity causes conflict and leaves rows unchanged.
   - Concurrent overlapping reservations do not oversell and do not partially persist on failure.
   - Basket validation sums duplicate product rows correctly for both stock and outlet quantities.

### Can be follow-up / do not block

- Localization wording polish.
- Typo cleanup like `inventoryProductAvailableQuantitiy`.
- DRYing the duplicated inventory/outlet allocation code, once the correct algorithm is settled.
- Frontend live clamp UX refinements.

### Final merge recommendation

**Request changes before merge.** The PR solves the original multi-row availability bug, but the raw-SQL allocation must be made atomic and its concurrency semantics clarified. The minimum safe merge bar is:

1. Transaction around the full deduction loop.
2. `.AsNoTracking()` or consistent EF update strategy.
3. Explicit handling of no-row and CAS-failure cases.
4. Confirm FIFO/LIFO.
5. Add at least core allocation tests.

After those are addressed, the remaining issues are mostly maintainability/contract-confirmation items rather than hard blockers.

dawiddworak88 (changes requested):
Almost there @PiotrZaranek 

# Code Review — Inventory Multi-Row Deduction & Outlet Quantity Clamping PR

## Summary
This PR refactors inventory and outlet quantity deduction to work across **multiple rows** per product (FIFO by `CreatedDate`), adds frontend clamping for outlet quantities, changes several domain models from nullable to non-nullable, fixes minor formatting issues, and adds new localization strings.

---

## Remarks (ranked by importance)

### 🔴 1. **BUG — Wrong `remainingToAllocate` guard in `InventoryService` (will silently under-deduct)**
**File:** `InventoryService.cs`, line 553

```csharp
if (remainingToAllocate < 0)  // ← BUG
    throw new ConflictException(...);
```

`remainingToAllocate` can **never go negative** with the current logic — it starts at `bookedQuantity` and is only reduced by `Math.Min(liveQuantity, remainingToAllocate)`, so the floor is 0. If concurrency causes some `affected == 0` (skipped rows), `remainingToAllocate` stays **positive**, not negative, and the exception is **never thrown**. The order will be placed but inventory will be **under-deducted** — a data-corruption bug.

Compare with `OutletService.cs` line 602 which correctly uses `> 0`:
```csharp
if (remainingToAllocate > 0)  // ← CORRECT
    throw new ConflictException(...);
```

**Fix:** Change `< 0` to `> 0` in `InventoryService`.

---

dawiddworak88 (approved):

SebaKas21 (approved):

PR #530, opened 2026-03-27
Basket expected lead time

SebaKas21:


dawiddworak88 (commented):
# Code Review: Expected Lead Time Feature

## Overview
This diff introduces an **Expected Lead Time** feature across the full stack — from API constants and database migrations through backend services, repositories, and controllers, all the way to the frontend React components. The feature allows tracking and displaying an expected delivery date (lead time) for order/basket items.

---

## 🔴 Critical Issues

### 1. **Commented-out DataProtection in Production Code** (`Ordering.Api/Program.cs`)
```csharp
/*
builder.Services.AddDataProtection().UseCryptographicAlgorithms(...)
    .PersistKeysToStackExchangeRedis(...);
*/
```
**This is a security concern.** DataProtection with Redis key persistence has been commented out. This appears to be a debugging/development leftover and **must not be merged**. Without DataProtection properly configured, the application may fail to decrypt cookies, tokens, or other protected payloads across restarts or in multi-instance deployments.

### 2. **Two TableCells on One Line — Missing `moreInfo` Column** (`StatusOrder.js`)
```jsx
<TableCell>{item.expectedLeadTime ? moment.utc(item.expectedLeadTime).local() : ""}</TableCell>                                                    <TableCell>{item.moreInfo}</TableCell>
```
Two `<TableCell>` elements are on the same line. While this technically works, the **header row added a new `expectedLeadTime` column but the `moreInfo` column cell is now squeezed onto the same line**. More critically, looking at the header, `expectedLeadTime` was added **after** `moreInfoLabel`, but in the body row, the `expectedLeadTime` cell **replaces** the position where `moreInfo` was — the `moreInfo` cell is still there but the column alignment is now **broken**. The header has: `...moreInfo | expectedLeadTime | unitPrice...` but the body has: `...expectedLeadTime + moreInfo (same cell position) | unitPrice...`. **This is a bug — columns will be misaligned.**

---

## 🟡 Significant Issues

### 3. **Inconsistent `moment()` Usage for `DateOnly` Display**
The `ExpectedLeadTime` is a `DateOnly?` on the backend (no time component). However, the frontend renders it inconsistently:
- `NewOrderForm.js`: `moment.utc(item.expectedLeadTime).local()` — converts UTC to local (unnecessary for a date-only value, and `.local()` returns a moment object, not a formatted string)
- `EditOrderForm.js`: `moment(item.expectedLeadTime).local()` — no `.utc()` first
- `StatusOrder.js`: `moment.utc(item.expectedLeadTime).local()` — same as NewOrderForm
- `OrderForm.js` (Seller): `moment(item.expectedLeadTime).local()` — no `.utc()`

**Issues:**
- `moment(...).local()` returns a moment object, not a string. You need `.format("L")` or similar to render a human-readable date.
- For a `DateOnly` value, UTC/local conversion is semantically wrong and can shift the date by ±1 day depending on timezone.
- **Recommendation:** Use `moment(item.expectedLeadTime).format("YYYY-MM-DD")` or a locale-aware format like `.format("L")` consistently.

### 4. **Buyer's `ProductsApiController` Missing `customerId` Parameter in Lead Time Call**
In `Buyer.Web/Areas/Products/ApiControllers/ProductsApiController.cs`:
```csharp
var leadTimes = await _leadTimeRepository.GetLeadTimesAsync(
    accessToken: token,
    skus: [.. products.Data.Select(x => x.Sku)]);
```
But the Buyer's `ILeadTimeRepository` interface (not shown in diff but implied) apparently accepts only `token` and `skus` — no `customerId`. Meanwhile, the Seller's version passes `customerId: client.OrganisationId`. 

The API endpoint used in Buyer is `LeadTimeBySkusApiEndpoint` (`/api/v1/product-lead-times/my`) while Seller uses `LeadTimeByCustomerApiEndpoint` (`/api/v1/product-lead-times`). This is intentional differentiation, but the Buyer's repository implementation is not shown in the diff — **ensure it exists and uses the correct endpoint**.

### 5. **`OrganisationId` Added to `Client` Domain Model Without Migration/Mapping Context**
In `Seller.Web/Areas/Clients/DomainModels/Client.cs`, `OrganisationId` was added as a non-nullable `Guid`:
```csharp
public Guid OrganisationId { get; set; }
```
If the API doesn't always return this field, it will default to `Guid.Empty`, which would then be passed to the lead time API. Ensure the client API always provides this value, or make it nullable with appropriate null checks.

### 6. **Potential N+1 / Performance Issue with Lead Time Lookups**
In both Buyer and Seller `ProductsApiController`, lead times are fetched and then matched with `FirstOrDefault` in a loop:
```csharp
LeadTimeDays = leadTimes?.Items?.FirstOrDefault(x => x.Sku == product.Sku)?.LeadTimeDays ?? 0
```
For large product lists, this is O(n×m). Consider converting to a dictionary lookup for O(1) access per product.

### 7. **Seller's `ExpectedDeliveryTime.js` is a Partial Copy Without `HolidaysConstants.js` Import Verification**
The new file `fe/projects/Seller.Portal/src/shared/components/ExpectedDeliveryTime/ExpectedDeliveryTime.js` imports from `./HolidaysConstants` — the `HolidaysConstants.js` file is also added in the diff, so this is fine. However, this is **duplicated code** from the AspNetCore project. Consider extracting to a shared package to avoid drift.

---

## 🟢 Minor Issues / Suggestions

### 8. **Localization Inconsistency**
- English: `"Expected date of lead time:"` — awkward phrasing. Consider `"Expected lead time:"` or `"Expected delivery date:"`
- Polish: `"Przewidywana data realizacji"` — missing colon (`:`) at the end, unlike DE and EN versions
- German: `"Voraussichtlicher Liefertermin:"` — has colon, good

### 9. **Missing Newline at End of File**
Multiple `.resx` files and `OrderFormViewModel.cs` (Seller) end with `</root>` or `}` without a trailing newline (`\ No newline at end of file`). While not a functional issue, it's a common best practice to include trailing newlines.

### 10. **`OrderPageModelBuilder` Locale Change**
```csharp
- Locale = CultureInfo.CurrentUICulture.Name,
+ Locale = componentModel.Language,
```
This change from `CultureInfo.CurrentUICulture.Name` to `componentModel.Language` seems intentional but is unrelated to the lead time feature. It should be called out separately or in its own commit to maintain clean git history.

### 11. **Removed `ILogger` and `Newtonsoft.Json` Import**
In `Buyer.Web/ProductsApiController.cs`, `ILogger<ProductsApiController>` and related imports were removed. In `BasketsController.cs`, `Newtonsoft.Json` was removed. Ensure these are truly unused and not needed elsewhere in the file.

### 12. **Hardcoded Holiday Dates**
The `HolidaysConstants.js` contains hardcoded Polish holidays through 2033. This is a maintenance burden — movable holidays (Easter-dependent) need manual updates. Consider computing Easter dates algorithmically or fetching holidays from a service.

### 13. **`calculateExpectedDeliveryDate` Business Logic on Frontend**
The expected delivery date calculation happens on the frontend (`useOrderManagement.js` and `OrderForm.js`). This means:
- Different clients could calculate different dates based on their local clock
- The business logic is duplicated between frontend and backend concerns
- Consider calculating this server-side for consistency

### 14. **Double Semicolon**
In `Seller.Web/ProductsApiController.cs`:
```csharp
var client = await _clientsRepository.GetClientAsync(token, _options.Value.DefaultCulture, clientId); ;
```
There's a double semicolon `;;` — harmless but sloppy.

---

## Summary

| Severity | Count | Key Items |
|----------|-------|-----------|
| 🔴 Critical | 2 | Commented-out DataProtection; Broken column alignment in StatusOrder |
| 🟡 Significant | 5 | Inconsistent date formatting; Missing customerId; Non-nullable OrganisationId; Performance; Code duplication |
| 🟢 Minor | 7 | Localization, newlines, unrelated changes, hardcoded holidays, double semicolon |

**Recommendation:** Do not merge until the critical issues (especially the DataProtection comment-out and the StatusOrder column misalignment) are resolved.

dawiddworak88 (approved):

PR #544, opened 2026-04-22
Team members

SebaKas21:


be/src/Project/Services/Ordering/Ordering.Api/Program.cs
```diff
@@ -80,13 +80,13 @@
         true);
 }
 
-builder.Services.AddDataProtection().UseCryptographicAlgorithms(
```
dawiddworak88, line 83: Shouldn't we bring it back?

be/src/Project/Web/Buyer/Buyer.Web/Areas/Orders/ModelBuilders/OrdersPageCatalogModelBuilder.cs
```diff
@@ -85,6 +86,11 @@ public async Task<CatalogOrderViewModel<Order>> BuildModelAsync(OrdersPageCompon
                         IsDateTime = false
                     },
                     new CatalogPropertyViewModel
+                    {
+                        Title = nameof(Order.CreatedBy).ToCamelCase(),
```
dawiddworak88, line 90: What does ToCamelCase help here with?

dawiddworak88 (changes requested):
## Code review — ranked by importance

### 🔴 Critical / blocker

1. **Client team member API is not tenant-safe; a seller can read/update/delete users from other organisations**
   - Files: `Identity.Api/v1/Controllers/ClientTeamMembersController.cs`, `Identity.Api/Services/ClientTeamMembers/ClientTeamMemberService.cs`
   - The new controller accepts `organisationId` from query/body instead of deriving it from `AccountConstants.Claims.OrganisationIdClaim`, unlike the existing `TeamMembersController`.
   - `Get(id)`, `Delete(id)` and `Update(id)` do not filter by organisation at all:
     - `FirstOrDefaultAsync(x => x.Id == model.Id.ToString())`
   - `Get(list)` lists accounts for any `organisationId` supplied by the caller.
   - `CreateAsync` reassigns an existing user to the requested organisation:
     - `user.OrganisationId = organisation.Id;`
   - Impact: cross-tenant data leak/account takeover/admin action across clients. This should block the PR.
   - Fix: derive organisation/client context server-side from authenticated claims or from a seller-owned client record; validate ownership; include `OrganisationId` predicate on every read/update/delete; never reassign an existing user across organisations without explicit validated flow.

2. **Existing client account / approval functionality was removed from the client edit flow**
   - Files: `Seller.Web/Areas/Clients/ApiControllers/ClientsApiController.cs`, `ClientFormModelBuilder.cs`, `ClientFormViewModel.cs`, `SaveClientRequestModel.cs`, `fe/.../ClientForm.js`
   - Removed behavior includes:
     - creating/updating/resetting client user account from the client form,
     - assigning roles based on client groups,
     - saving client/user approvals.
   - This is not only a UI refactor; `ClientsApiController` no longer calls `_identityRepository.UpdateAsync`, `_identityRepository.AssignRolesAsync`, or `_userApprovalsRepository.SaveAsync` when `HasAccount` is set.
   - Order placement relevance: buyer checkout checks user approvals in `Buyer.Web/Areas/Orders/ApiControllers/BasketCheckoutApiController.cs` and uses approval `ToSendOrderConfirmationEmails` to set `HasApprovalToSendEmail`. Removing the old approval assignment path can silently change order-confirmation email behavior for client users. Order creation may still succeed, but existing account/approval workflows are broken.
   - Fix: either preserve the old client account flow until fully migrated, or provide a backward-compatible migration path and ensure approvals/roles are handled for the new client team member flow.

3. **Ordering service production data protection was commented out**
   - File: `Ordering.Api/Program.cs`
   - The PR comments out Redis-backed `AddDataProtection().PersistKeysToStackExchangeRedis(...)`.
   - This is unrelated to client team members / `CreatedBy` and is risky in production. If anything in Ordering depends on ASP.NET Core data protection now or later, keys become instance-local/ephemeral, which can break encrypted payloads across restarts/replicas.
   - Order placement relevance: Ordering is part of the checkout event pipeline. Even if current bearer-token/event handling does not directly need this, this is a high-risk infrastructure regression in the order service and should not be included casually.
   - Fix: revert this change unless there is a separate, documented infrastructure migration.

### 🟠 High

4. **Client team member edit URL passes the client id as `organisationId`**
   - File: `Seller.Web/Areas/Clients/ModelBuilders/ClientTeamMemberPageCatalogModelBuilder.cs`
   - `EditUrl` is built with:
     - `clientId = client?.Id`
     - `organisationId = componentModel.Id`
   - In this page, `componentModel.Id` is the client id, not the organisation id. `NewUrl` correctly uses `client.OrganisationId`, but `EditUrl` does not.
   - Because the Identity update service currently ignores organisation ownership, this may appear to work, but it masks the tenant-security bug above and can break once proper validation is added.
   - Fix: use `organisationId = client?.OrganisationId`.

5. **New client team member pages can null-reference when client id is missing or invalid**
   - File: `ClientTeamMemberPageCatalogModelBuilder.cs`
   - If `componentModel.Id` is missing or `_clientsRepository.GetClientAsync(...)` returns null, the code still dereferences `client.OrganisationId` / `client.Id` for `NewUrl`, `SearchApiUrl`, and initial `PagedItems`.
   - Impact: direct navigation, stale links, or deleted clients can crash the page.
   - Fix: handle null client explicitly: return 404/no-content view, disable new/search URLs, or throw controlled exception.

6. **Client team member creation ignores requested disabled state**
   - File: `Identity.Api/Services/ClientTeamMembers/ClientTeamMemberService.cs`
   - `CreateClientTeamMemberServiceModel` has `IsDisabled`, but new `ApplicationUser` is always created with `IsDisabled = false`.
   - Impact: seller may create an inactive invited member in UI, but backend activates them anyway.
   - Fix: set `IsDisabled = model.IsDisabled` or remove the field from the create UI.

7. **Saving user approvals after team member save is not transactional**
   - File: `Seller.Web/Areas/Clients/ApiControllers/ClientTeamMembersApiController.cs`
   - Flow:
     1. create/update identity user,
     2. then save approvals.
   - If approvals save fails, the user has already been created/updated and UI returns an error. Retrying can cause confusing partial state.
   - Fix: preferably expose a backend operation that saves member + approvals atomically, or implement compensating/clear retry behavior.

### 🟡 Medium

8. **`CreatedBy` for orders may often be null / unreliable**
   - Files: `Basket.Api/v1/Controllers/BasketsController.cs`, `Ordering.Api/...`
   - `CreatedBy` is populated from `ClaimTypes.Name`:
     - `User.Claims.FirstOrDefault(x => x.Type == ClaimTypes.Name)?.Value`
   - Existing checkout code elsewhere commonly uses `ClaimTypes.Email`; token/profile code appears to consistently include email and organisation id, but `ClaimTypes.Name` may not always be present.
   - Order placement impact: this should not fail order creation because `CreatedBy` is nullable and not validated, but the new UI column may be empty for many orders.
   - Fix: decide expected value and source it consistently, e.g. full name with fallback to email.

9. **Client team member list/search does not search by email**
   - File: `ClientTeamMemberService.cs`
   - Search filters first name, last name, or exact id only. The UI sorts/displays email and likely users will search by email.
   - Fix: include `Email.StartsWith(model.SearchTerm)` or a case-insensitive contains strategy consistent with the rest of the app.

10. **Client team member form lacks fetch error handling**
   - File: `fe/projects/Seller.Portal/src/areas/Clients/components/ClientTeamMemberForm/ClientTeamMemberForm.js`
   - Unlike `ClientForm.js`, the new form does not `.catch(...)`. Network/runtime errors can leave loading state stuck and no toast shown.
   - Fix: add catch/finally to reset loading and show `generalErrorMessage`.

11. **Drawer menu was not updated with the new Client Team Members section**
   - Files: `Seller.Web/Shared/ModelBuilders/MenuTiles/MenuTilesModelBuilder.cs`, `DrawerMenuModelBuilder.cs`
   - Menu tiles include the new feature, drawer menu does not. Depending on layout, users may not discover it from the side navigation.
   - Fix: add localized drawer item too, using `TeamMembersResources` or global localization consistently.

### 🟢 Lower / cleanup

12. **Several unused imports/constructor dependencies remain after moving account logic**
   - Examples: `ClientsApiController.cs`, `ClientFormModelBuilder.cs`, `ClientTeamMemberPageCatalogModelBuilder.cs`
   - Not functionally blocking, but it obscures the diff and may trigger analyzer warnings.

13. **`OrderServiceModel` property reorder is harmless but noisy**
   - File: `Ordering.Api/ServicesModels/OrderServiceModel.cs`
   - `Attachments` moved before timestamps while adding `CreatedBy`; no issue, just noise.

## Order placement assessment

The `CreatedBy` propagation itself is mostly backward-compatible: the event property is optional, the new DB column is nullable, and checkout validators were not tightened. So I do **not** see a direct failure where adding `CreatedBy` prevents an order from being placed.

However, I would **not merge this PR as-is** because it introduces severe tenant-security issues in client team members, removes existing client account/approval behavior that can affect order-confirmation email permissions, and includes an unrelated risky change to Ordering API data protection.

dawiddworak88 (changes requested):
## Code review of `diff-output.txt` — ranked remarks

### 1. **Critical — Client team member endpoints allow cross-organisation user access/modification**
**Files:**
- `be/src/Project/Services/Identity/Identity.Api/v1/Controllers/ClientTeamMembersController.cs`
- `be/src/Project/Services/Identity/Identity.Api/Services/ClientTeamMembers/ClientTeamMemberService.cs`

The new `ClientTeamMembersController` accepts `organisationId` from query/body and passes it through directly. Unlike the existing `TeamMembersController`, it does **not** derive organisation scope from `AccountConstants.Claims.OrganisationIdClaim`.

In addition, several service methods do not enforce organisation scoping:

```csharp
_context.Accounts.FirstOrDefaultAsync(x => x.Id == model.Id.ToString())
```

This appears in `GetAsync`, `DeleteAsync`, and `UpdateAsync`. As a result, any authenticated caller who can guess/obtain a user id may read, update, disable, or delete users outside their organisation. `CreateAsync` also looks up users by email globally and can reassign `user.OrganisationId` to the supplied organisation.

**Recommendation:**
- Do not trust `organisationId` from request body/query for authorization.
- Extract organisation id from claims in the controller, as `TeamMembersController` does.
- Enforce `x.OrganisationId == model.OrganisationId` in all read/update/delete queries.
- Be especially careful with existing-user-by-email flows; do not silently move users between organisations.

---

### 2. **High — `ClientTeamMemberPageCatalogModelBuilder` generates edit URLs with the wrong organisation id**
**File:** `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ModelBuilders/ClientTeamMemberPageCatalogModelBuilder.cs`

This line appears inconsistent:

```csharp
viewModel.EditUrl = _linkGenerator.GetPathByAction("Edit", "ClientTeamMemberDetail", new { Area = "Clients", culture = CultureInfo.CurrentUICulture.Name, clientId = client?.Id, organisationId = componentModel.Id });
```

`componentModel.Id` is the **client id**, not the organisation id. The surrounding code uses `client.OrganisationId` for `NewUrl`, `SearchApiUrl`, and data loading.

This can cause edits to post back with a client id as `organisationId`. Currently the backend does not enforce organisation scope, which masks this bug. Once the security issue above is fixed, this will likely break updates.

**Recommendation:** use `organisationId = client?.OrganisationId`.

---

### 3. **High — `ClientTeamMemberPageCatalogModelBuilder` can throw `NullReferenceException` when client is missing**
**File:** `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ModelBuilders/ClientTeamMemberPageCatalogModelBuilder.cs`

`client` is nullable, but later accessed unconditionally:

```csharp
organisationId = client.OrganisationId
clientId = client.Id
...
viewModel.PagedItems = await _clientTeamMembersRepository.GetAsync(..., client.OrganisationId, ...)
```

If `componentModel.Id` is missing, invalid, deleted, or inaccessible, this page will fail with a 500 instead of a controlled not-found/no-content flow.

**Recommendation:** handle `client is null` explicitly: return a not-found model, redirect, or avoid building URLs/data that require `client`.

---

### 4. **High — Existing client account creation/update behavior was removed from `ClientsApiController`**
**Files:**
- `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ApiControllers/ClientsApiController.cs`
- `fe/projects/Seller.Portal/src/areas/Clients/components/ClientForm/ClientForm.js`
- `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ViewModels/ClientFormViewModel.cs`

The diff removes the logic that previously:
- created/updated the identity user when `model.HasAccount` was set,
- assigned roles based on client groups,
- saved user/client approvals.

The frontend also removes the create/reset account button and approval switches from `ClientForm`.

If this is an intentional product change moving account management to “Client Team Members”, it should be explicitly confirmed and migrated. Otherwise, this is a functional regression: saving a client with account-related fields will no longer create/update the buyer account or role/approval data.

**Recommendation:** either restore the behavior or document/complete the migration path so existing workflows and data are not silently lost.

---

### 5. **Medium — `CreateClientTeamMemberServiceModel.IsDisabled` is ignored**
**File:** `be/src/Project/Services/Identity/Identity.Api/Services/ClientTeamMembers/ClientTeamMemberService.cs`

`CreateAsync` accepts `model.IsDisabled`, and the UI exposes an active/inactive switch, but new users are always created active:

```csharp
IsDisabled = false
```

For existing users, `CreateAsync` also updates name/organisation but does not apply `model.IsDisabled`.

This differs from the updated `TeamMemberService`, which now uses `model.IsDisabled` on create and existing-user update.

**Recommendation:** apply `model.IsDisabled` consistently, or remove the switch from the create flow if disabled-on-create is unsupported.

---

### 6. **Medium — Partial failure possible when saving client team member approvals**
**File:** `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ApiControllers/ClientTeamMembersApiController.cs`

The API first saves the identity user, then saves approvals:

```csharp
teamMemberId = await _clientTeamMembersRepository.SaveAsync(...);
await _userApprovalsRepository.SaveAsync(..., teamMemberId.Value);
```

If approval saving fails, the catch returns 500, but the user may already have been created/updated. This leaves the system in a partially saved state.

**Recommendation:** handle this explicitly. For example, return a more specific partial-failure message, retry approvals, or design the backend operation so identity user and approval assignment are coordinated more safely.

---

### 7. **Medium — Client team member search does not include email**
**File:** `be/src/Project/Services/Identity/Identity.Api/Services/ClientTeamMembers/ClientTeamMemberService.cs`

The new client team member search only checks first name, last name, and id:

```csharp
x.FirstName.StartsWith(model.SearchTerm)
|| x.LastName.StartsWith(model.SearchTerm)
|| x.Id.ToString() == model.SearchTerm
```

The updated existing `TeamMemberService` now includes email search. Since the UI displays and sorts by email, users will likely expect email search to work here too.

**Recommendation:** add `x.Email.StartsWith(model.SearchTerm)` for parity.

---

### 8. **Medium — `ClientTeamMemberForm` mutates approval props directly**
**File:** `fe/projects/Seller.Portal/src/areas/Clients/components/ClientTeamMemberForm/ClientTeamMemberForm.js`

The handler mutates the `approval` object received from props:

```javascript
approval.isApproved = !approval.isApproved;
```

This is an anti-pattern in React and can produce hard-to-debug rendering/state issues. The old client form had similar logic, but this is newly introduced for the client team member form.

**Recommendation:** keep approvals in local state or derive checked state entirely from `teamMemberApprovalIds`.

---

dawiddworak88 (changes requested):
Hi @SebaKas21, here remarks after review - at the bottom detailed plan to address point 1.

## Code review remarks ranked by importance

### 1. **Critical — client-team-member endpoints allow cross-tenant access/modification**
**Files:**
- `be/src/Project/Services/Identity/Identity.Api/v1/Controllers/ClientTeamMembersController.cs`
- `be/src/Project/Services/Identity/Identity.Api/Services/ClientTeamMembers/ClientTeamMemberService.cs`

The new `ClientTeamMembersController` accepts `organisationId` from the request/query body and only checks `IsSeller()`. Unlike the existing `TeamMembersController`, it does **not** derive/validate organisation ownership from authenticated claims.

Examples:
- `Get(..., Guid? organisationId)` passes arbitrary `organisationId` into `GetClientTeamMembersServiceModel`.
- `Save(...)` passes `request.OrganisationId` into create/update.
- `Delete(Guid? id)` and `Get(Guid? id)` do not pass or validate organisation ownership at all.
- Service methods `GetAsync(GetClientTeamMemberServiceModel)`, `DeleteAsync(...)`, and `UpdateAsync(...)` query by `Id` only.

This means any seller who can call the endpoint may enumerate, edit, disable, or delete users belonging to another client organisation if they know/guess ids or organisation ids.

**Recommendation:**
- Never trust `OrganisationId` from client input for authorization.
- Resolve seller identity from claims.
- Verify that the seller is allowed to manage the target client organisation, probably through seller-client relationship data.
- Add `OrganisationId` filtering to single-item `Get`, `Update`, and `Delete` queries, similar to existing `TeamMemberService` patterns.

---

### 2. **Critical — deleting a client team member physically removes an `ApplicationUser` account**
**File:** `be/src/Project/Services/Identity/Identity.Api/Services/ClientTeamMembers/ClientTeamMemberService.cs`

`DeleteAsync` does:

```csharp
_context.Accounts.Remove(user);
await _context.SaveChangesAsync();
```

This is very dangerous for identity data. A team member may have approvals, claims, roles, audit history, orders, comments, or other foreign-key-related records. Hard deletion may break referential integrity or remove auditability.

The risk is amplified by the missing tenant validation from finding #1.

**Recommendation:**
Use a soft-delete/disable model (`IsDisabled = true`, possibly `IsActive = false` if available), remove role/approval associations explicitly if required, and enforce tenant ownership before any destructive operation.

---

### 3. **High — update flow can disable/update the wrong user by email and does not validate email uniqueness/change semantics**
**Files:**
- `be/src/Project/Services/Identity/Identity.Api/Services/ClientTeamMembers/ClientTeamMemberService.cs`
- Existing similar issue in `TeamMemberService.UpdateAsync`

`UpdateAsync` loads the target user by `Id`, then separately loads another user by `model.Email`:

```csharp
var existingUserWithEmail = await _context.Users.FirstOrDefaultAsync(x => x.Email == model.Email);
if (existingUserWithEmail is not null)
{
    existingUserWithEmail.IsDisabled = model.IsDisabled;
}
```

Problems:
- If `model.Email` belongs to a different account, this mutates that other account’s `IsDisabled` flag.
- `user.Email` itself is not updated, so accepting `Email` in the update model is misleading.
- There is no conflict handling if email change is intended.
- Query uses `_context.Users` while other logic uses `_context.Accounts`; if these are aliases it is still confusing, and if not, it may mutate an unexpected set.

**Recommendation:**
- For update by id, modify only the loaded `user`.
- If email changes are unsupported, validate that `model.Email == user.Email` or remove email from update payload.
- If email changes are supported, implement explicit uniqueness checks and update `Email`, `NormalizedEmail`, `UserName`, and `NormalizedUserName` consistently.

---

### 4. **High — existing client account/approval creation was removed from client save, possibly causing regression**
**Files:**
- `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ApiControllers/ClientsApiController.cs`
- `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ModelBuilders/ClientFormModelBuilder.cs`
- `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ViewModels/ClientFormViewModel.cs`
- `fe/projects/Seller.Portal/src/areas/Clients/components/ClientForm/ClientForm.js`

The diff removes the `HasAccount` / account creation / reset-password / client approvals flow from the regular client form and save API. This may be intentional if replaced by dedicated client team members, but it changes existing behavior significantly:

- `ClientsApiController.Index` no longer calls identity update, role assignment, or user approval saving.
- `ClientForm.js` no longer exposes the “create account/reset password” action.
- `ClientApprovalIds` was removed from `SaveClientRequestModel`.

If existing users rely on creating a primary client account from the client edit page, this is a functional regression.

**Recommendation:**
Confirm product requirements. If client accounts are now managed exclusively through “Client Team Members”, provide migration/backward-compatible UX and ensure the first/default client account can still be created and managed.

---

### 5. **High — client-team-member page model builder has likely null-reference failures and incorrect route values**
**File:** `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ModelBuilders/ClientTeamMemberPageCatalogModelBuilder.cs`

`client` may remain null when `componentModel.Id` is missing or `_clientsRepository.GetClientAsync(...)` returns null, but the code then dereferences it:

```csharp
organisationId = client.OrganisationId
clientId = client.Id
...
organisationId = client.OrganisationId
...
GetAsync(..., client.OrganisationId, ...)
```

Also this route looks suspicious:

```csharp
viewModel.EditUrl = ... new { clientId = client?.Id, organisationId = componentModel.Id }
```

`organisationId` is set to `componentModel.Id`, which appears to be the client id, not the client organisation id. New URL uses `client.OrganisationId`, edit URL uses `componentModel.Id`; this inconsistency can break edit/save because the API POST relies on `organisationId` query parameter.

**Recommendation:**
- Return a safe error or redirect when client is missing.
- Use `client.OrganisationId` consistently for `organisationId`.
- Add tests for navigating from client list → team member list → edit/new.

---

### 6. **Medium — `ClientTeamMembersApiController.Post` catches all exceptions and returns 500, hiding validation/authorization errors**
**File:** `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ApiControllers/ClientTeamMembersApiController.cs`

The broad catch:

```csharp
catch
{
    return StatusCode(500, ...);
}
```

turns identity API validation failures, conflicts, forbidden responses, and not-found cases into generic 500 errors. It also loses diagnostic context.

**Recommendation:**
Catch known `CustomException`/API exceptions and return their actual status/message. Let unexpected exceptions propagate to centralized exception handling/logging.

---

### 7. **Medium — creating a client team member reassigns any existing user with matching email to another organisation**
**File:** `be/src/Project/Services/Identity/Identity.Api/Services/ClientTeamMembers/ClientTeamMemberService.cs`

`CreateAsync` does:

```csharp
var user = await _context.Accounts.FirstOrDefaultAsync(x => x.Email == model.Email);
if (user is not null)
{
    user.OrganisationId = organisation.Id;
    ...
}
```

If the email already belongs to a user in a different organisation, this silently moves that user to the selected client organisation. That can break access for the existing organisation and is a serious data integrity issue.

**Recommendation:**
If a user exists with the same email but a different organisation, return conflict unless explicit cross-organisation reassignment is a supported, audited operation.

---

### 8. **Medium — new Identity API list endpoint lacks service-model validation**
**File:** `be/src/Project/Services/Identity/Identity.Api/v1/Controllers/ClientTeamMembersController.cs`

A `GetClientTeamMembersModelValidator` was added but is never used in the list `Get(...)` action. This allows missing `organisationId`, invalid paging values, or other invalid base model values to reach the service.

**Recommendation:**
Use the validator consistently, and add explicit rules for `OrganisationId`, `PageIndex`, and `ItemsPerPage` if required.

---

### 9. **Low/Medium — frontend approval timestamp label appears never populated**
**Files:**
- `fe/projects/Seller.Portal/src/areas/Clients/components/ClientTeamMemberForm/ClientTeamMemberForm.js`
- `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ModelBuilders/ClientTeamMemberFormModelBuilder.cs`
- `be/src/Project/Web/Seller/Seller.Web/Areas/Clients/ViewModels/ClientTeamMemberFormViewModel.cs`

The React component renders:

```jsx
{props.expressedOnLabel}: ...
```

But `ClientTeamMemberFormViewModel` has no `ExpressedOnLabel`, and the model builder does not set it. The UI will display `undefined: date`.

**Recommendation:**
Add `ExpressedOnLabel` to the view model and populate it, or remove the label from the component.

---

### 10. **Low — several unused imports / injected dependencies will generate warnings and reduce maintainability**
Examples:
- `ClientsApiController.cs`: unused imports such as `Foundation.GenericRepository.Paginations`, `Seller.Web.Areas.Clients.Repositories.Groups`, `Seller.Web.Areas.Shared.Repositories.UserApprovals`, `Seller.Web.Shared.Repositories.Identity`, `System.Collections.Generic`, `System.Linq.Dynamic.Core`; constructor still accepts `IClientGroupsRepository clientGroupsRepository` but does not use it.
- `ClientFormModelBuilder.cs`: still injects `IIdentityRepository`, `IApprovalsRepository`, `IUserApprovalsRepository` but removed usage.
- `ClientTeamMemberPageCatalogModelBuilder.cs`: `DocumentFormat.OpenXml.Office2010.Excel` appears unrelated.
- `ClientTeamMembersPageCatalogModelBuilder.cs`: several unused usings.

**Recommendation:**
Clean up unused dependencies/usings and remove DI parameters no longer needed.

---
---
**Plan for addressing point 1**


## Security Issue Analysis

### Root cause
The new `ClientTeamMembersController` (in `Identity.Api`) and the matching BFF controller (`Seller.Web`) trust an `organisationId` value that comes from the client (query string / request body / route id) and only check `IsSeller()`. Concretely:

- `GET /api/v1/clientteammembers?organisationId={X}` → service does `where u.OrganisationId == model.OrganisationId` with **no check that `X` belongs to the calling seller's tenant**.
- `POST /api/v1/clientteammembers` → `CreateAsync` looks up the organisation by `model.OrganisationId` and writes the user under it. A seller can move/create a user inside any organisation in the system, including another seller's organisation.
- `GET /api/v1/clientteammembers/{id}` and `DELETE /api/v1/clientteammembers/{id}` → service queries by user id only (`x.Id == model.Id.ToString()`), no organisation/tenant filter at all → cross-tenant read & delete (compare to `TeamMemberService.DeleteAsync` / `GetAsync` which require `OrganisationId == model.OrganisationId`).
- `UpdateClientTeamMemberServiceModel` accepts `OrganisationId`, but `ClientTeamMemberService.UpdateAsync` ignores it for the lookup — same IDOR.
- BFF `ClientTeamMembersApiController` (`Seller.Web`) reads `organisationId` from query string / form and forwards it without checking that this organisation actually belongs to a `Client` that this seller owns (Clients table has `SellerId` = seller-org id).

The result: any authenticated user with the `Seller` role can list, read, modify, create or delete team members for **any client of any seller** in the system. This is a classic multi-tenant IDOR — confidentiality + integrity break across tenants.

### Authoritative ownership facts (already in the codebase)
- `Identity.Api.Organisations` contains both seller and buyer organisations, but does **not** record which seller owns which buyer org.
- The seller↔buyer link lives in `Client.Api`: `Client.SellerId` (= seller's organisation id) → `Client.OrganisationId` (= buyer organisation id, which is what becomes `ApplicationUser.OrganisationId`).
- Therefore tenant ownership of a buyer organisation must be verified by going through Client.Api (e.g. via `IClientsRepository` already used in `Seller.Web`, or via a new internal endpoint exposed for Identity.Api).
- The existing `TeamMembersController` (sellers' own employees) gets the `OrganisationId` from `AccountConstants.Claims.OrganisationIdClaim` and uses it as the `OrganisationId` filter — that is the pattern the new controller violates.

---

## Phase-by-Phase Remediation Plan

### Phase 0 — Containment / triage (immediate)
**Goal:** stop the bleed before refactoring.

1. **Step 0.1** — Open a security ticket; flag the affected endpoints in CHANGELOG/security advisory.
2. **Step 0.2** — Hot-patch (smallest safe change to merge today):
   - In `ClientTeamMembersController.Save(...)` and `Get(... organisationId)`: override the `OrganisationId` coming from the request with `null` and reject the request with `403` until Phase 1 ships, OR
   - Temporarily disable route registration for `ClientTeamMembersController` (comment out `services.AddScoped<IClientTeamMemberService, ClientTeamMemberService>();` plus remove the controller route attribute) if the feature is not yet in production.
3. **Step 0.3** — Audit logs: search Identity & Seller.Web request logs for any `organisationId` values that do not match the caller's seller-org claim. Any anomalies → invalidate sessions + notify.

### Phase 1 — Identity.Api hardening (server-side authoritative fix)
**Goal:** every code path must derive the *seller's* organisation from the JWT and verify that the target buyer-organisation/team-member belongs to that seller. Never trust client-supplied org/user ids alone.

**File: `Identity.Api/v1/Controllers/ClientTeamMembersController.cs`**

1. **Step 1.1** — Replace `IsSeller()` with a strongly-typed check that also extracts the seller organisation id:
   ```csharp
   var sellerClaim = User.Claims.FirstOrDefault(x => x.Type == AccountConstants.Claims.OrganisationIdClaim);
   var sellerOrganisationId = GuidHelper.ParseNullable(sellerClaim?.Value);
   if (!IsSeller() || sellerOrganisationId is null) return StatusCode((int)HttpStatusCode.Forbidden);
   ```
   (Mirror the pattern used in `TeamMembersController`.)
2. **Step 1.2** — Stop accepting `organisationId` from request models for *write* operations on an existing record:
   - `Delete(Guid? id)`, `Get(Guid? id)`: drop any reliance on caller-supplied org id; pass `sellerOrganisationId` to the service, which enforces ownership.
   - `Save(...)` (Update branch): ignore `request.OrganisationId`; resolve the existing user, then enforce that that user's `OrganisationId` belongs to a client owned by `sellerOrganisationId`.
   - `Save(...)` (Create branch) and `Get(... organisationId, ...)`: still need a target buyer-org id, but it must be validated against `sellerOrganisationId`.
3. **Step 1.3** — Pass `SellerOrganisationId` (from claim) through the service models:
   - Add `Guid? SellerOrganisationId` to `Get/Update/Create/Delete/GetClientTeamMemberServiceModel` and `GetClientTeamMembersServiceModel`.
   - Update all matching validators (`Validators/ClientTeamMembers/*`) to require it (`NotEmpty().NotNull()`).

**File: `Identity.Api/Services/ClientTeamMembers/ClientTeamMemberService.cs`**

4. **Step 1.4** — Introduce a dependency that can answer “does buyer-org `X` belong to seller-org `Y`?”:
   - Option A (preferred, minimum coupling): introduce `IClientOwnershipValidator` in Identity.Api that calls `Client.Api`'s existing `clients/organisation/{organisationId}` (or a new admin endpoint `clients/owner/{organisationId}` that returns `SellerId`) using a service-to-service token.
   - Option B (if cross-service HTTP is undesirable): mirror the seller↔buyer mapping into Identity.Api by consuming a `ClientCreatedIntegrationEvent` / `ClientUpdatedIntegrationEvent` and storing `SellerId` on the `Organisation` row (add a migration `AddedSellerIdToOrganisations`). Then ownership is a single EF query.
   - Document the choice in the PR. Either way the API surface is `Task<bool> BelongsToSellerAsync(Guid buyerOrganisationId, Guid sellerOrganisationId)`.
5. **Step 1.5** — Apply ownership filter consistently in every method:
   - `CreateAsync`: after loading `organisation`, call `BelongsToSellerAsync(model.OrganisationId, model.SellerOrganisationId)`. If false → throw `CustomException(... Forbidden)`.
   - `UpdateAsync`: load the user, then `BelongsToSellerAsync(user.OrganisationId, model.SellerOrganisationId)`. Reject if false. Also reject any attempt to *change* the OrganisationId (don't write `user.OrganisationId = ...`).
   - `DeleteAsync`: change query to `x.Id == model.Id.ToString()`, then validate ownership the same way before `Remove`. (This matches the existing `TeamMemberService.DeleteAsync` pattern that filters by `OrganisationId`.)
   - `GetAsync(GetClientTeamMemberServiceModel)`: same ownership check after the user is loaded.
   - `GetAsync(GetClientTeamMembersServiceModel)`: the list query becomes `where u.OrganisationId == model.OrganisationId AND <model.OrganisationId belongs to model.SellerOrganisationId>`. If `OrganisationId` is null, restrict to all client-orgs owned by the seller.
6. **Step 1.6** — Bug fix in `UpdateAsync`: it currently mutates an unrelated `_context.Users` row that happens to share an email (`existingUserWithEmail.IsDisabled = model.IsDisabled`). Remove that block — it is both a bug and a privilege-escalation vector (you can disable arbitrary users by registering an email collision). Replace with normal `user.*` updates only.
7. **Step 1.7** — Tighten `IClientTeamMemberService` and `ClientTeamMemberServiceModel` types to match.

**Validators**

8. **Step 1.8** — Update each validator under `Validators/ClientTeamMembers` to require `SellerOrganisationId` and to forbid clients from supplying it (set it from claim only at the controller layer; validators just assert it is present).

### Phase 2 — Seller.Web BFF hardening (defense in depth)
**Goal:** BFF should not blindly forward client input.

**File: `Seller.Web/Areas/Clients/ApiControllers/ClientTeamMembersApiController.cs`**

1. **Step 2.1** — Inject `IClientsRepository` and resolve the seller's `Client` for the supplied `organisationId` before any call to Identity.Api:
   - `var client = await _clientsRepository.GetByOrganisationIdForSellerAsync(token, language, organisationId);`
   - If `client` is null (or its `SellerId` ≠ caller's seller-org claim) → return `403`.
2. **Step 2.2** — For `Delete(id)` and any future `GetById`: do not accept `organisationId` from the URL/body. Instead, look the team member up first (call Identity.Api `GET /clientteammembers/{id}` which now enforces tenancy) or look up the client by walking from the user back to its organisation via Client.Api.
3. **Step 2.3** — Remove the (now unused) acceptance of `organisationId` in `Post(...)` query string; take it only from `model.OrganisationId` server-side, validated against the caller.
4. **Step 2.4** — Add a `[Authorize]` attribute / role-based authorization filter that requires the `Seller` role (right now the controller has no `[Authorize]`). Consider an authorization policy `SellerOwnsClientOrganisation` that wraps the check above.
5. **Step 2.5** — Audit `IClientsRepository` / `ClientsRepository.GetByOrganisationAsync` (introduced in this diff) to ensure it filters by the calling seller's `SellerId` server-side; if not, fix Client.Api accordingly.

### Phase 3 — Client.Api support (if Option A in Step 1.4)
1. **Step 3.1** — Add an internal endpoint `GET /api/v1/clients/owner?organisationId={id}` returning `{ sellerId }`, locked behind a dedicated machine-to-machine scope (`identity.client.read`), so Identity.Api can validate ownership without reusing user tokens.
2. **Step 3.2** — Document the new scope in the IdentityServer client config + helm charts (`misc/deployment/IdentityApi/helm-chart/values.yaml`).
3. **Step 3.3** — Add a small typed client + Polly resilience around the call.

---

dawiddworak88 (approved):

PR #548, opened 2026-05-05
backend dependency upgrade All Services

MarcinKalinowski1:


dawiddworak88 (changes requested):
# Code Review: .NET 10 Upgrade + Centralized Package Management

This PR is largely a platform/tooling upgrade: migration from `.sln` → `.slnx`, .NET 8 → .NET 10, `Microsoft.AspNetCore.Mvc.Versioning` → `Asp.Versioning.*`, `Microsoft.Azure.KeyVault` → `Azure.Security.KeyVault.Secrets`, introduction of Central Package Management (`Directory.Packages.props`) and a common `Directory.Build.props`. There are also a few unrelated functional changes (LeadTime pagination, removal of Foundation.EventLog, ExpectedDeliveryTime calc fix).

Below are remarks ordered by importance.

---

## 🔴 Critical / Likely Bugs

### 1. `SalesFact.cs` — `Currency` becomes non‑nullable but column is still nullable
```csharp
[Column(TypeName = "nvarchar(3)")]
public string Currency { get; set; }   // was: string?
```
The DB column likely allows nulls and existing rows may have NULL currencies. With nullable reference types enabled this becomes a `string` (non‑nullable), which:
- May produce CS8618 warnings.
- More importantly, the property is propagated through `OrderItem`, `OrderItemStartedEventModel`, `CreateSalesAnalyticsProductServiceModel`, `SalesService.CreateSalesFact(..., string currency)`. Old messages on the bus (still containing `null`) will deserialize OK but will start hitting unintended code paths. Combined with `RelationalEventId.PendingModelChangesWarning` being suppressed (see #4) the schema/model mismatch is now silently tolerated.

Either keep `string?` or run a migration making it `NOT NULL` with a default, and validate inbound integration events.

### 2. `SalesService.CreateSalesFact` — `IsStock`/`IsOutlet` are still swapped
```csharp
IsOutlet = isStock,
IsStock = isOutlet,
```
This is pre‑existing, but the file was touched in this PR. Fix it now or the analytics rollup is wrong by definition.

### 3. `LeadTimeRepository.GetLeadTimesAsync` — NRE on failed response, wrong page size, dropped contract change
```csharp
if (!response.IsSuccessStatusCode || response.Data?.Items is null)
{
    _logger.LogError(...);   // logs but does NOT return / continue
}
leadTimeResults.AddRange(response.Data.Items);   // NRE if Data is null
```
- Missing `continue;` (or `return`) after logging — guaranteed `NullReferenceException` on a failed page.
- `Constants.MaxItemsPerPage` is used for both batching SKUs *and* the paging passed to the API. Those are conceptually different. If the API returns page size N and you sent N SKUs, you may still be missing items if duplicates/filters change the count. Loop until `pagedSkus.Length` equals returned items, or until `Items.Count() < requested`.
- Return type changed from `PagedLeadTimeResults` → `IEnumerable<LeadTimeItem>`. Callers using `result?.Items?.FirstOrDefault(...)` were updated to `result?.FirstOrDefault(...)` — but several files still showed `Items` access in the diff being removed; double‑check every consumer (Buyer.Web has 8+ call sites, Seller.Web likely too) for compilation and behavior parity.
- `if (skus.Length == 0) return default;` returns `null` for `IEnumerable<>`. Prefer `Array.Empty<LeadTimeItem>()` so callers don't need null checks (and many of the diff sites already assume non‑null via `?.FirstOrDefault`, which is fine, but inconsistent with the empty‑collection idiom).

### 4. Globally suppressing `PendingModelChangesWarning`
Every `AddDbContext` call now has:
```csharp
.ConfigureWarnings(w => w.Ignore(RelationalEventId.PendingModelChangesWarning))
```
This was added in EF 9/10 specifically to alert you that your model is out of sync with the last migration. Suppressing it everywhere hides genuine schema drift caused by the .NET 10 upgrade (e.g., new DateOnly mappings, identity column changes, the `Currency` nullability above). Recommendation: generate a migration, apply it, and remove the suppression — or at minimum suppress it only where you can prove the difference is benign (e.g., via a comment).

### 5. `IdentityServer4` signing‑credential rewrite — production‑critical regression risks
```csharp
var segments = secretUri.AbsolutePath.Split('/', StringSplitOptions.RemoveEmptyEntries);
var name = segments[1];
var version = segments.Length > 2 ? segments[2] : null;
```
- `segments[1]` assumes the path is `/secrets/{name}[/{version}]`. If the URI is malformed or KeyVault changes URL shape (it shouldn't, but) you get an `IndexOutOfRangeException` that bubbles up at app startup. Add a guard / proper parser (`KeyVaultSecretIdentifier` from `Azure.Security.KeyVault.Secrets` does exactly this).
- Synchronous `.Value` (`secretClient.GetSecret(...).Value`) and `X509CertificateLoader.LoadPkcs12(...)` are called on the startup thread — fine for this codepath, but no retry/transient fault handling around the KeyVault call. The previous code didn't either, but this is the moment to add Polly/`Azure.Core` retry.
- `X509CertificateLoader.LoadPkcs12` is .NET 9+. Confirm runtime version of the IdentityServer host actually matches `Directory.Build.props` (`net10.0`) — otherwise you get a `MissingMethodException` only in production.
- No `using` / disposal of the `X509Certificate2`. Long‑lived signing creds are typically OK, but document the lifetime expectation.
- Missing config validation: if any of `tenantId`/`clientId`/`clientSecret`/`certificateSecretIdentifier` is null, `ClientSecretCredential` throws `ArgumentNullException` at runtime. Check upfront and fail fast with a clear message.


### 5. Removal of `Foundation.EventLog` project
The whole project (api request models, definitions, repositories, DI registration) is deleted. The diff doesn't show callers being updated — verify with a grep that no service still calls `RegisterEventLogDependencies()` or `IEventLogRepository`. If any leftover `using Foundation.EventLog` exists, the build will fail; if it's only unused references in csproj/sln, fine. Worth a quick repo‑wide search.

---

## 🟠 High — Likely Issues / Strong Smells

### 7. CI workflows reference `AspNetCore.slnx`
```yaml
API_SOLUTION_PATH: '${{ github.workspace }}/be/src/AspNetCore.slnx'
```
- `dotnet build`/`test` support `.slnx` only from SDK 9.0.200+. The repo pins `global.json` to `10.0.203`, so this is fine for the current runner — but the `azure-build-validation-pipeline.yml` doesn't have a `UseDotNet@2` step shown. Make sure the Azure pipeline image has SDK ≥ 9.0.200 (Windows‑hosted images sometimes lag).
- `azure-build-validation-pipeline.yml` was updated to `.slnx` but it doesn't seem to set up the SDK. Ensure pipeline has `UseDotNet@2 version: 10.0.203` (or `useGlobalJson: true`).

### 8. `Buyer.Web` `BuyerHeaderViewModel` lost `SearchLabel` and `SearchPlaceholderLabel`
```csharp
- public string SearchLabel { get; set; }
- public string SearchPlaceholderLabel { get; set; }
```
If the Razor view (`_Header.cshtml`) or React component still references these properties, you'll get a runtime `NullReferenceException` (Razor) or `undefined` in JS. Verify these were intentionally removed and consumers are updated. Same for `MainNavigation` removed from `OrdersPageViewModel` and the empty `PagedDeliveryAddressesByIdsRequestModel` (now just inherits — make sure callers no longer set `Ids`).

### 9. `OrdersPageViewModel` — removed `MainNavigation`
```csharp
- public MainNavigationViewModel MainNavigation { get; set; }
```
Verify the orders page Razor / React no longer reads it. This is one of those "compiles fine, breaks at runtime" changes.

### 10. `PagedDeliveryAddressesByIdsRequestModel` — removed `Ids`
```csharp
public class PagedDeliveryAddressesByIdsRequestModel : PagedDeliveryAddressesRequestModel
{
-    public string Ids { get; set; }
}
```
This is a public API contract. If clients still POST `Ids`, model binding will silently drop it. Confirm the parent class now exposes `Ids`, otherwise the endpoint behavior changed.

---

## 🟡 Medium — Maintainability / Hygiene

### 11. Lots of unrelated noise commits — BOM removals
Almost every csproj diff includes `-<Project Sdk=...>` → `<Project Sdk=...>` (BOM stripped). Combined with the .NET 10 + CPM changes this hides actual content. Two suggestions:
- Add a `.editorconfig` rule (`charset = utf-8` not `utf-8-bom`) and run `dotnet format` once.
- Split this PR into: (a) BOM/format normalization, (b) .NET 10 + CPM, (c) Asp.Versioning rename, (d) KeyVault rewrite, (e) functional fixes (LeadTime pagination, ExpectedDeliveryTime, view model cleanups). Reviewing all together is much harder.

### 12. `slnx` — multiple projects pinned to `Project="Debug"` for `NoVariablesDefault`
```xml
<BuildType Solution="NoVariablesDefault|*" Project="Debug" />
```
For Analytics, DownloadCenter, Global, Tests, Telemetry. Looks accidental — most other projects map `NoVariablesDefault → Release`. Either intended (shipping debug symbols for these) or a mistake when migrating from the old `.sln`. Worth a quick check.

### 13. `Catalog.Api` removed `AspNetCore.Mvc.Versioning` 5.1.0 ✓ but added `Azure.Identity` and `Microsoft.Identity.Client` — what for?
The Catalog.Api code itself doesn't appear to need these. Likely required transitively by something (Grula SDK? KeyVault?). If it's not used directly, leave it but add a comment in Directory.Packages.props, or remove. Same applies to `Catalog.BackgroundTasks`, `Client.Api`, `Inventory.Api`, `Media.Api`, `News.Api`, `Ordering.Api`. CPM means the ItemGroup references serve as documentation; keep them tight.

### 14. `ExpectedDeliveryTime.js` — fix is correct, but tests?
```js
const deliveryDate = moment(expectedDeliveryDate).startOf("day");
const now = moment().startOf("day");
const days = deliveryDate.diff(now, "days");
```
This is the right fix (avoids off‑by‑one when comparing partial days). If any unit/snapshot test asserted the old "X days from now" calculation it'll need updating. Also: `moment` is in maintenance mode; consider `dayjs`/`date-fns` as a follow‑up.

### 15. Integration tests — Testcontainers 4.x major migration
The `IImage` interface gained `Registry`, `Digest`, `Platform`, `MatchLatestOrNightly`, `MatchVersion(...)`. The wrappers (`BasketApiImage`, `CatalogApiImage`, etc.) have been updated identically. Two issues:
- 7 nearly identical wrapper classes — extract a base `DotNetServiceImage` or use a single generic one parameterized by Dockerfile path.
- `WithWaitStrategy(...UntilPortIsAvailable(8080))` → `UntilInternalTcpPortIsAvailable(8080)`. Same semantics, but be aware `UntilInternalTcpPort...` does the check inside the container, so it requires `nc`/`bash` in the image. Slim ASP.NET base images typically have neither — verify your tests still wait correctly. Often `UntilHttpRequestIsSucceeded(...)` against `/health` is more robust.
---

PR #549, opened 2026-05-07
frontend dependency upgrade Account, Buyer, Seller

MarcinKalinowski1:


dawiddworak88 (changes requested):
## Code review

### 1. **High — React-Toastify styles were removed globally, but `ToastContainer` is still used everywhere**

**Files:**
- `fe/projects/Account/src/shared/layouts/index.scss`
- `fe/projects/AspNetCore/src/shared/layouts/index.scss`
- `fe/projects/Seller.Portal/src/shared/layouts/index.scss`
- e.g. `fe/projects/Account/src/areas/Accounts/pages/SignIn/SignInPage.js:7-12`

The diff removes `@import "~react-toastify/dist/ReactToastify.css";` from shared layout styles, but pages still render `<ToastContainer />` and code still calls `toast.success/error/...` extensively.

React-Toastify still requires its CSS unless the app intentionally provides complete custom styling. Without it, notifications can render unstyled or incorrectly positioned.

**Recommendation:** restore the CSS import once per app, e.g. in each shared layout entrypoint, or import it from a top-level JS entry that is guaranteed to load on every page using toasts.

---

### 2. **High — MUI X DatePicker migration likely passes incompatible value types**

**Files:**
- `fe/projects/AspNetCore/src/shared/components/AnalyticsDashboard/SalesAnalytics/SalesAnalytics.js:25-36`, `100-132`
- `fe/projects/Seller.Portal/src/shared/components/AnalyticsDashboard/SalesAnalytics/SalesAnalytics.js:25-36`, `100-132`
- similar Seller analytics components: `ClientSalesAnalytics`, `CountrySalesAnalytics`, `ProductsSalesAnalytics`
- `fe/projects/Seller.Portal/src/areas/Inventory/components/InventoryForm/InventoryForm.js:28`, `69`
- `fe/projects/Seller.Portal/src/shared/helpers/validators/ChartValidator.js:3-5`

The migration replaces old picker components with `@mui/x-date-pickers` + `AdapterMoment`, but state is initialized directly from server props:

```js
const [fromDate, setFromDate] = useState(props.fromDate);
const [toDate, setToDate] = useState(props.toDate);
```

With `AdapterMoment`, `DatePicker.value` should be a `moment` instance or `null`, not an arbitrary string/`Date`. The handlers also send the picker value directly through `JSON.stringify`, and `ChartValidator` compares these values with `new Date()`.

This can cause invalid picker values, runtime warnings/errors, incorrect comparisons, or changed request payload shape.

**Recommendation:** normalize boundary values explicitly, for example:

```js
const [fromDate, setFromDate] = useState(props.fromDate ? moment(props.fromDate) : null);
const [toDate, setToDate] = useState(props.toDate ? moment(props.toDate) : null);
```

Then serialize intentionally before API calls, e.g. `date?.toISOString()` or the backend’s expected date format, and update `ChartValidator` to compare normalized values.

---

### 3. **Medium — MUI `ListItem button` migration lost button semantics/accessibility**

**Files:**
- `fe/projects/Account/src/shared/components/Header/Header.js`
- `fe/projects/Seller.Portal/src/shared/components/Header/Header.js:61-75`

The diff removes the deprecated/removed `button` prop from `ListItem`:

```jsx
<ListItem onClick={handleDrawerClose}>
...
<ListItem selected={item.isActive}>
```

This avoids the MUI API issue, but it also removes the intended interactive behavior: focus handling, keyboard accessibility, ripple/hover behavior, and proper button/list semantics. The menu item case also wraps `ListItem` in an `<a>`, which is less ideal than making the MUI component render as the anchor.

**Recommendation:** migrate to `ListItemButton`, e.g.:

```jsx
<ListItemButton onClick={handleDrawerClose}>
  ...
</ListItemButton>

<ListItemButton component="a" href={item.url} selected={item.isActive}>
  ...
</ListItemButton>
```

---

### 4. **Medium — GitHub Action major-version bumps should be verified/pinned before merging**

**Files:**
- `.github/workflows/giuru-*-ci.yaml`
- `.github/workflows/giuru-*-cd-*.yaml`

The workflows bump multiple actions to new major tags, for example:

```yaml
uses: actions/checkout@v6
uses: actions/setup-node@v6
uses: docker/setup-buildx-action@v4
uses: docker/login-action@v4
uses: docker/build-push-action@v7
```

If any of these tags are not published, the workflow fails before build/deploy. Even when published, broad major upgrades can include breaking input/runtime behavior.

**Recommendation:** verify every major tag exists and review each migration guide. For release pipelines, consider pinning to known-good major/minor versions or SHAs rather than speculative newest majors.

---

---

## Summary

Most of the migration direction looks reasonable, but I would address the Toastify CSS removal and DatePicker value normalization before merging. I would also verify the GitHub Actions major tags and use `ListItemButton` for the header drawer migration.

PR #553, opened 2026-05-19
Use client's OrganisationId as SellerId when admin places order on behalf of client

SebaKas21:


dawiddworak88 (changes requested):
## Code review:

### Summary
The PR is aiming at the right root cause: orders created from the seller panel were persisted with `Order.SellerId` equal to the seller/admin organisation claim, while buyer-side order listing filters non-seller users by `Order.SellerId == current client organisation id`.

The new flow sends `client?.OrganisationId` from Seller Web into Basket API as `SellerId`, and Basket API publishes it into the ordering event. Functionally, this should make seller-created orders visible to the client **when the selected client is resolved correctly**.

However, I would **not approve this as-is** because the Basket API now trusts a caller-supplied organisation identifier that controls order visibility.

---

## Findings

### 1. High severity — Basket API trusts caller-supplied `SellerId`, allowing order visibility manipulation

**Files:**
- `be/src/Project/Services/Basket/Basket.Api/v1/Controllers/BasketsController.cs:220-225`
- `be/src/Project/Services/Basket/Basket.Api/Services/BasketService.cs:33-40`
- `be/src/Project/Services/Basket/Basket.Api/v1/RequestModels/BasketCheckoutRequestModel.cs`

The PR adds `SellerId` to the public checkout request and maps it directly into the integration event:

```csharp
SellerId = request.SellerId,
```

Then:

```csharp
SellerId = checkoutBasketServiceModel.SellerId ?? checkoutBasketServiceModel.OrganisationId,
```

This means any authenticated caller of Basket API can influence `Order.SellerId`, which is the field used later for buyer visibility:

```csharp
if (model.IsSeller is false)
{
    orders = orders.Where(x => x.SellerId == model.OrganisationId.Value);
}
```

Even if the normal Buyer Web UI does not send `SellerId`, direct API callers can. A malicious or buggy client could create an order visible to another client organisation, or hide the order from themselves.

**Recommendation:**
Do not trust `SellerId` from the request body. Prefer one of these approaches:

1. Rename the concept to something like `ClientOrganisationId` / `BuyerOrganisationId` to reflect how it is actually used for visibility.
2. Compute it server-side from `ClientId`, ideally in Basket API or Ordering API, not from caller input.
3. If Basket API must accept it temporarily, validate all of the following:
   - the caller is a seller/admin when `SellerId` is supplied;
   - `SellerId` equals the organisation id of `ClientId`;
   - the seller/admin has access to that client.
4. For non-seller buyer checkout, ignore `request.SellerId` entirely and always use the authenticated user organisation claim.

At minimum, the Basket controller should do something like:

```csharp
SellerId = isSellerClaim is not null
    ? resolvedClientOrganisationId
    : GuidHelper.ParseNullable(sellerClaim?.Value)
```

—not accept this value from the request body.

---

### 2. Medium severity — New `SellerId` request field in Seller Web is unused and misleading

**File:**
- `be/src/Project/Web/Seller/Seller.Web/Areas/Orders/ApiRequestModels/CheckoutBasketRequestModel.cs:7-10`

The PR adds:

```csharp
public Guid? SellerId { get; set; }
```

But `BasketCheckoutApiController` does not use `model.SellerId`; it derives the value from:

```csharp
client?.OrganisationId
```

That is good from a trust perspective at the Seller Web boundary, but exposing this request field is confusing and invites future misuse.

**Recommendation:**
Remove `SellerId` from `Seller.Web/Areas/Orders/ApiRequestModels/CheckoutBasketRequestModel.cs` unless the frontend actually needs it. The server already derives the correct value.

---

### 3. Medium severity — Basket checkout validation was not updated for the new visibility-critical field

**File:**
- `be/src/Project/Services/Basket/Basket.Api/ServicesModelsValidators/CheckoutBasketServiceModelValidator.cs:11-16`

The Basket service model validator still only requires:

```csharp
OrganisationId
ClientId
ClientName
BasketId
Username
```

For the seller-created order case, `SellerId` / client organisation id is now the key piece of data needed for buyer visibility. If `request.SellerId` is missing or invalid, BasketService silently falls back to `OrganisationId`, which recreates the original bug for seller-panel orders.

**Recommendation:**
Make the rule explicit:

- for buyer checkout: use authenticated `OrganisationId`;
- for seller/admin checkout: require a resolved client organisation id and validate it against `ClientId`.

Avoid silent fallback for seller checkout, because fallback hides the failure and causes the same visibility bug.

---

### 4. Medium severity — Seller checkout can still fail with `NullReferenceException` when client lookup returns null

**File:**
- `be/src/Project/Web/Seller/Seller.Web/Areas/Orders/ApiControllers/BasketCheckoutApiController.cs:53-70`

The controller checks `client is not null` for approval lookup, but later uses:

```csharp
client.Email,
```

If `_clientsRepository.GetClientAsync(...)` returns null, checkout fails with a null reference before reaching Basket API.

This is not introduced by the PR, but it is in the exact code path being changed and should be handled while fixing this area.

**Recommendation:**
Return a controlled validation/not-found response when the selected client cannot be resolved, e.g. before the repository call:

```csharp
if (client is null)
{
    // return 422 or 404 with localized message
}
```

---

### 5. Low/medium severity — Lead time changes are unrelated and partially defensive, but Buyer implementation still has a null path

**Files:**
- `be/src/Project/Web/Buyer/Buyer.Web/Shared/Repositories/LeadTime/LeadTimeRepository.cs:34-83`
- `be/src/Project/Web/Seller/Seller.Web/Shared/Repositories/LeadTime/LeadTimeRepository.cs:32-70`

The added `LeadTimeUrl` guard is sensible if the service is optional. Most consumers use `leadTimes?.FirstOrDefault(...)`, so returning `default` is unlikely to break display logic.

However, the Buyer repository still logs an error and then dereferences `response.Data.Items` even when `response.Data` or `Items` is null:

```csharp
if (!response.IsSuccessStatusCode || response.Data?.Items is null)
{
    _logger.LogError(...);
}

leadTimeResults.AddRange(response.Data.Items);
```

That can still throw.

**Recommendation:**
After logging, `continue` or `return default`:

```csharp
if (!response.IsSuccessStatusCode || response.Data?.Items is null)
{
    _logger.LogError(...);
    continue;
}

leadTimeResults.AddRange(response.Data.Items);
```

Also, this change is unrelated to the order visibility bug and would ideally be in a separate PR unless it is required by deployment configuration.

---

## Additional existing concern discovered during review

### Existing security gap — Get order by id does not enforce organisation visibility

**Files:**
- `be/src/Project/Services/Ordering/Ordering.Api/Services/OrdersService.cs:291-294`
- `be/src/Project/Services/Ordering/Ordering.Api/v1/Controllers/OrdersController.cs:322-340`

List endpoints filter non-seller users by organisation, but `GetAsync(GetOrderServiceModel)` fetches by id only:

```csharp
var existingOrder = await _context.Orders.FirstOrDefaultAsync(x => x.Id == model.Id && x.IsActive);
```

So a client who knows an order id may be able to fetch an order outside their organisation. This is not introduced by the PR, but it is directly related to order visibility and access control.

**Recommendation:**
Apply the same organisation filter to single-order reads and dependent endpoints.

---

## Recommended tests before merge

Add coverage for the visibility contract:

1. **Buyer checkout** without `SellerId` creates an order whose visibility organisation equals the authenticated buyer organisation claim.
2. **Seller checkout for a client** creates an order whose visibility organisation equals `client.OrganisationId`, not the seller organisation claim.
3. **Buyer checkout with malicious `SellerId` in request body** is ignored or rejected.
4. **Seller checkout with mismatched client/organisation id** is rejected.
5. **Client order list** includes seller-created orders for that client.
6. **Client order detail** rejects orders outside the client organisation.

---

## Verdict

**Request changes.**

The functional idea is correct, but the implementation should not expose and trust `SellerId` as a request-body value. The value that controls client order visibility should be derived and validated server-side, preferably from the selected `ClientId` and authenticated caller context.

dawiddworak88 (approved):

PR #554, opened 2026-05-25
Optimize product indexing to eliminate timeouts

SebaKas21:


dawiddworak88 (changes requested):
Hi @SebaKas21, let's have a look at the following issues, below proposals how to address them: 

### 1. **High — Stale Elasticsearch documents are no longer removed**
**File:** `be/src/Infrastructure/Catalog/Foundation.Catalog/Repositories/ProductIndexingRepositories/BulkProductIndexingRepository.cs`  
**Lines:** `151-196`

The PR removes the previous `DeleteByQueryAsync` step before indexing and now only sends `Index` operations for documents that can currently be built.

That means old documents can remain in Elasticsearch when a product no longer produces a document for all previously indexed cases, for example:

- product loses all active translations,
- supported cultures change,
- a translation becomes inactive,
- a product ID is passed for reindexing but the product is no longer returned by the DB query,
- category/schema/translation data causes a document not to be emitted.

Because search queries rely on indexed documents, those stale docs can continue appearing until a full seller delete/rebuild happens.

**Recommendation:** keep a delete/update strategy per product before indexing, or explicitly delete missing language documents. If using deterministic IDs like `$"{ProductId}_{Language}"`, delete by IDs or by `ProductId` before reindexing the current product set.

---

### 2. **High — Product attributes are skipped when the category schema is missing for the exact language**
**File:** `BulkProductIndexingRepository.cs`  
**Lines:** `57-75`, `277-291`

The old code had a fallback:

```csharp
FirstOrDefaultAsync(x => x.CategoryId == categoryId && x.Language == language && x.IsActive)
    ?? FirstOrDefaultAsync(x => x.CategoryId == categoryId && x.IsActive)
```

The new code preloads schemas, but `PopulateProductAttributes` only looks up:

```csharp
schemaCaches.TryGetValue((categoryId, language), out var schemaObject)
```

If a category only has a schema in one language, or a product translation falls back to another active translation, product attributes will now be silently omitted for other supported cultures.

This is risky because `ProductAttributes` are used downstream for size, palette, color, price group, packing, etc.

**Recommendation:** preserve the fallback behavior when building the cache, for example by resolving `(categoryId, language)` to an exact schema first and then any active schema for the category.

---

### 3. **Medium — Removing the final Elasticsearch refresh changes observable behavior after indexing**
**File:** `BulkProductIndexingRepository.cs`  
**Lines:** `77-80`, `181-187`

The previous implementation called:

```csharp
await _elasticClient.Indices.RefreshAsync(Indices.Index<ProductSearchModel>());
```

The new implementation removes that and explicitly sets:

```csharp
.RefreshOnCompleted(false)
```

After `IndexAsync` / `IndexBatchAsync` completes, callers may still read stale search results until Elasticsearch’s refresh interval runs. That is a behavior change, especially for single-product indexing flows where the caller likely expects the awaited operation to be searchable immediately.

**Recommendation:** either restore a final refresh for interactive/single-product indexing, or make the delayed consistency explicit and ensure all callers tolerate it. A possible compromise is refreshing only once after the whole batch/rebuild, not per bulk page.

---

### 4. **Medium — Bulk indexing failures are not propagated and the wait result is ignored**
**File:** `BulkProductIndexingRepository.cs`  
**Lines:** `198-206`

The code checks invalid bulk items only by logging:

```csharp
_logger.LogError("Some bulk indexing items failed in batch.");
```

Then it returns `Task.CompletedTask` regardless. Also, `waitHandle` is assigned but never inspected. As a result, `ProductsService.IndexAllAsync` can log “Completed bulk indexing” even when some documents failed or the bulk operation timed out.

**Recommendation:** include failed item details in the log and fail the task on unrecoverable bulk errors/timeouts, or at minimum return a status that callers can use. Also remove or check the unused `waitHandle` variable.

---

## Minor notes

- `ProcessBatchAsync` is no longer actually async; it blocks synchronously on `BulkAll(...).Wait(...)` and returns `Task.CompletedTask`. This is not necessarily incorrect in a background task, but the method name/signature now hides synchronous blocking behavior.
- The logger type is still `ILogger<ProductIndexingRepository>` inside `BulkProductIndexingRepository`; existing issue, but worth cleaning up while touching this file.

---

### Fix 1 — Stale documents (High)
**Problem (line 77-80):** The old code deleted existing ES documents for the product IDs before re-indexing. The new code only upserts documents it can build, so if a product loses its translations or becomes unindexable, its old ES document persists.

**Fix:** Add a delete-by-product-IDs step at the start of `IndexBatchAsync`, before the batch loop. Since you use deterministic doc IDs (`{ProductId}_{Language}`), a targeted delete is straightforward:

```csharp
public async Task IndexBatchAsync(IEnumerable<Guid> productIds)
{
    var supportedCultures = _configuration["SupportedCultures"].Split(",");
    var productIdsList = productIds.ToList();

    // ✅ Delete existing documents for these products before re-indexing
    if (productIdsList.Any())
    {
        await _elasticClient.DeleteByQueryAsync<ProductSearchModel>(d => d
            .Query(q => q.Terms(t => t.Field(f => f.ProductId).Terms(productIdsList))));
    }

    // ... rest of the method unchanged
```

This is lightweight (one query, scoped to the batch) and ensures stale documents are cleaned up before re-indexing.

---

### Fix 2 — Missing schema fallback (High)
**Problem (lines 64-75, 287-291):** The old code had a fallback: if no schema matched `(categoryId, language)`, it would try `(categoryId, anyActiveLanguage)`. The new preloaded cache only stores exact `(categoryId, language)` keys, so products in categories without a schema for every supported culture silently lose their `ProductAttributes`.

**Fix:** Add a fallback lookup in `PopulateProductAttributes`:

```csharp
private void PopulateProductAttributes(
    ProductSearchModel document, 
    string formData, 
    Guid categoryId, 
    string language,
    Dictionary<(Guid categoryId, string language), JObject> schemaCaches)
{
    if (string.IsNullOrWhiteSpace(formData))
        return;

    var schemaKey = (categoryId, language);
    if (!schemaCaches.TryGetValue(schemaKey, out var schemaObject))
    {
        // ✅ Fallback: try any schema for this category
        schemaObject = schemaCaches
            .Where(kv => kv.Key.categoryId == categoryId)
            .Select(kv => kv.Value)
            .FirstOrDefault();

        if (schemaObject == null)
            return;
    }

    // ... rest unchanged
```

Alternatively, when building `schemaCaches` in `IndexBatchAsync`, you can pre-populate fallback keys for every supported culture:

```csharp
// After the initial cache build loop, add fallback entries
foreach (var categoryId in categoryIds)
{
    // Find the first available schema for this category
    var fallback = schemaCaches
        .Where(kv => kv.Key.categoryId == categoryId)
        .Select(kv => kv.Value)
        .FirstOrDefault();

    if (fallback != null)
    {
        foreach (var culture in supportedCultures)
        {
            var key = (categoryId, culture);
            if (!schemaCaches.ContainsKey(key))
            {
                schemaCaches[key] = fallback;
            }
        }
    }
}
```

I recommend the second approach — it keeps `PopulateProductAttributes` simple and does the fallback resolution once upfront.

---

### Fix 3 — Missing index refresh (Medium)
**Problem (line 186):** `RefreshOnCompleted(false)` + removal of `RefreshAsync` means newly indexed documents may not be searchable for up to 1 second (ES default refresh interval). This is fine for bulk rebuilds but problematic for single-product indexing via `IndexAsync(Guid productId)`.

**Fix:** Add a refresh after the batch loop in `IndexBatchAsync`:

```csharp
public async Task IndexBatchAsync(IEnumerable<Guid> productIds)
{
    // ... existing code ...

    await foreach (var batch in GetProductBatchesAsync(productIdsList))
    {
        await ProcessBatchAsync(batch, supportedCultures, schemaCaches);
    }

    // ✅ Refresh index so documents are immediately searchable
    await _elasticClient.Indices.RefreshAsync(Indices.Index<ProductSearchModel>());
}
```

This is one refresh call at the end of the entire operation, not per batch page — so it's efficient for bulk rebuilds too.

---

### Fix 4 — Fire-and-forget BulkAll + ignored failures (Medium)
**Problem (lines 198-206):** `BulkAll(...).Wait(...)` is called synchronously (blocking), but the method returns `Task.CompletedTask` regardless of whether the wait completed or items failed. The `waitHandle` return value is never checked.

**Fix:** Make `ProcessBatchAsync` properly async and propagate errors:

```csharp
private async Task ProcessBatchAsync(
    ProductBatchDto[] batch, 
    string[] supportedCultures, 
    Dictionary<(Guid categoryId, string language), JObject> schemaCaches)
{
    var documents = new List<ProductSearchModel>();

    // ... document building loop unchanged ...

    if (!documents.Any())
        return;

    var bulkAllObservable = _elasticClient.BulkAll(documents, b => b
        .Index<ProductSearchModel>()
        .BackOffRetries(3)
        .BackOffTime("5s")
        .Size(1000)
        .RefreshOnCompleted(false)
        .MaxDegreeOfParallelism(2)
        .BufferToBulk((descriptor, buffer) =>
        {
            foreach (var doc in buffer)
            {
                descriptor.Index<ProductSearchModel>(i => i
                    .Id($"{doc.ProductId}_{doc.Language}")
                    .Document(doc));
            }
        }));

    // ✅ Use Wait and check errors properly
    var failedItems = new List<string>();

    bulkAllObservable.Wait(TimeSpan.FromMinutes(10), onNext: response =>
    {
        var errors = response.Items.Where(i => !i.IsValid).ToList();
        if (errors.Any())
        {
            foreach (var error in errors)
            {
                failedItems.Add($"Id={error.Id}, Error={error.Error?.Reason}");
            }
        }
    });

    if (failedItems.Any())
    {
        _logger.LogError("Bulk indexing had {Count} failures: {Details}", 
            failedItems.Count, string.Join("; ", failedItems.Take(10)));
    }
}
```

Key changes:
- Method signature is `async Task` again (or at minimum not returning `Task.CompletedTask` after a blocking `.Wait()`)
- Failed items are collected with details for diagnostics
- Consider throwing if failure count exceeds a threshold

---

### Minor fix — Logger type
Line 20 uses `ILogger<ProductIndexingRepository>` instead of `ILogger<BulkProductIndexingRepository>`. Quick fix:

```csharp
private readonly ILogger<BulkProductIndexingRepository> _logger;
```

---

### Summary of changes needed

| # | Severity | What | Where |
|---|----------|------|-------|
| 1 | 🔴 High | Add `DeleteByQueryAsync` before re-indexing | `IndexBatchAsync`, before batch loop |
| 2 | 🔴 High | Add schema language fallback | `IndexBatchAsync` cache build or `PopulateProductAttributes` |
| 3 | 🟡 Medium | Add `RefreshAsync` after batch loop | End of `IndexBatchAsync` |
| 4 | 🟡 Medium | Check `BulkAll` failures, log details, don't return `Task.CompletedTask` | `ProcessBatchAsync` |
| 5 | 🟢 Low | Fix logger type parameter | Constructor |

dawiddworak88 (approved):
🎖️

PR #555, opened 2026-05-26
fix role claim mapping after IdentityModel upgrade

MarcinKalinowski1:


SebaKas21 (approved):

dawiddworak88 (changes requested):
## Code review

I reviewed the diff plus the surrounding authentication/claims code paths in:

### 1. **Potentially breaking role claim type mismatch in `Identity.Api` access-token validation**

**Severity:** High / blocking unless validated by tests

**File:** `be/src/Project/Services/Identity/Identity.Api/DependencyInjection/CompositionRoot.cs`

The PR changes Identity API token validation to:

```csharp
options.RoleClaimType = ClaimTypes.Role;
options.NameClaimType = ClaimTypes.Name;
```

But this same solution emits seller roles as `JwtClaimTypes.Role` (`"role"`), not `ClaimTypes.Role`:

- `be/src/Project/Services/Identity/Identity.Api/Areas/Accounts/Services/ProfileServices/ProfileService.cs`

```csharp
context.IssuedClaims.Add(new Claim(JwtClaimTypes.Role, AccountConstants.Roles.Seller));
```

- `be/src/Project/Services/Identity/Identity.Api/Services/Tokens/TokenService.cs`

```csharp
claims.Add(new Claim(JwtClaimTypes.Role, AccountConstants.Roles.Seller));
```

Also, IdentityServer resources expose `JwtClaimTypes.Role`:

- `be/src/Project/Services/Identity/Identity.Api/Areas/Accounts/Configurations/IdentityServerConfig.cs`

```csharp
new IdentityResource(AccountConstants.IdentityResources.Roles, new[] { JwtClaimTypes.Role })
...
JwtClaimTypes.Role
```

This is particularly suspicious because the existing client OIDC setup in `Foundation.Account` explicitly disables inbound mapping and uses `JwtClaimTypes.Role`:

```csharp
JwtSecurityTokenHandler.DefaultMapInboundClaims = false;
JwtSecurityTokenHandler.DefaultInboundClaimTypeMap.Clear();
...
RoleClaimType = JwtClaimTypes.Role
```

So the codebase currently has two conventions:

- Web/OIDC client auth: role claim type is `JwtClaimTypes.Role` (`"role"`)
- PR-changed API auth: role claim type is `ClaimTypes.Role` (`http://schemas.microsoft.com/ws/2008/06/identity/claims/role`)

For plain `AddJwtBearer`, this may work if inbound claim mapping remains enabled and maps `"role"` to `ClaimTypes.Role`. But `IdentityServer4.AccessTokenValidation` has historically defaulted around JWT claim names (`JwtClaimTypes.Role`) and may not behave identically. The PR makes `Identity.Api` depend on implicit mapping that is not obvious from the code and contradicts how tokens are issued.

**Recommendation:** Make the claim-type convention explicit and consistent. Either:

**1. Keep API validation role claim type as `JwtClaimTypes.Role` and change the new `RegisterApiAccountDependencies` config accordingly, or**

2. Add regression tests proving that tokens containing `"role": "Seller"` make `User.IsInRole(AccountConstants.Roles.Seller)` true in both:
   - APIs using `RegisterApiAccountDependencies`
   - `Identity.Api` using `AddIdentityServerAuthentication`

Without such a test, this PR could silently make `IsSeller` false for Identity API endpoints.

---

dawiddworak88 (approved):

PR #558, opened 2026-06-09
Add validation for stock and outlet quantity in seller order form

PiotrZaranek:


dawiddworak88 (changes requested):
## Code review remarks ranked by importance

### 1. **Critical / Blocking — checkout validation can use inventory/outlet quantities from other sellers**
**Files:**
- `be/src/Project/Web/Seller/Seller.Web/Areas/Orders/Services/Basket/BasketService.cs`
- `be/src/Project/Services/Inventory/Inventory.Api/Services/InventoryItems/InventoryService.cs`
- `be/src/Project/Services/Inventory/Inventory.Api/Services/OutletItems/OutletService.cs`

`BasketService.ValidateStockOutletQuantitiesAsync()` validates availability by calling:
- `_inventoryRepository.GetInventoryProductByProductIdsAsync(...)`
- `_outletRepository.GetOutletProductsByProductsIdAsync(...)`

However the inventory/outlet service queries used by those endpoints filter only by product id and active flags, not by the current seller/organisation:

```csharp
where model.Ids.Contains(i.ProductId) && product.IsActive && i.IsActive
```

and:

```csharp
.Where(x => model.Ids.Contains(x.ProductId) && x.Product.IsActive && x.IsActive)
```

That means availability may be summed across all sellers. A seller could pass checkout validation because another seller has stock/outlet availability for the same product id.

**Impact:** incorrect order acceptance / stock validation bypass in multi-tenant scenarios.

**Recommendation:** pass `OrganisationId`/seller id through these endpoints and filter by `SellerId == model.OrganisationId`. Alternatively expose seller-scoped inventory/outlet availability endpoints specifically for checkout validation.

---

### 2. **Critical / Blocking — `BasketNotFound` validation likely never works for missing baskets**
**Files:**
- `be/src/Project/Web/Seller/Seller.Web/Areas/Orders/Services/Basket/BasketService.cs`
- `be/src/Project/Services/Basket/Basket.Api/Services/BasketService.cs`
- `be/src/Project/Services/Basket/Basket.Api/v1/Controllers/BasketsController.cs`

The PR adds:

```csharp
if (basket is null) throw new CustomException(_orderLocalizer.GetString("BasketNotFound"), ...);
```

But `Basket.Api`’s `GetBasketById()` returns an empty basket object when the repository returns `null`:

```csharp
if (basket == null)
{
    return new BasketServiceModel
    {
        Id = serviceModel.Id.Value,
        Items = Array.Empty<BasketItemServiceModel>()
    };
}
```

So `Seller.Web` receives a non-null `Basket` with empty items, and `BasketNotFound` is not thrown. Checkout can then continue with a missing/empty basket.

**Impact:** the new validation does not enforce the intended “basket exists” condition; orders may be accepted without a real basket.

**Recommendation:** either make `Basket.Api` return `404/NoContent` for a missing basket, or make Seller.Web treat an empty basket returned from this endpoint as invalid for checkout. For example:

```csharp
if (basket?.Items == null || !basket.Items.Any())
{
    throw new CustomException(_orderLocalizer.GetString("BasketNotFound").Value, (int)HttpStatusCode.NotFound);
}
```

If empty baskets are valid elsewhere, consider a dedicated checkout validation endpoint instead of reusing the generic basket `GET` endpoint.

---

### 3. **High — pre-check has a race condition and does not guarantee checkout consistency**
**Files:**
- `be/src/Project/Web/Seller/Seller.Web/Areas/Orders/Services/Basket/BasketService.cs`
- `be/src/Project/Services/Basket/Basket.Api/Services/BasketService.cs`

The PR validates inventory before calling `CheckoutBasketAsync`, but the actual stock/outlet deduction happens later via basket checkout integration events. Between the Seller.Web validation and Basket.Api checkout/event handling, available quantities may change.

**Impact:** two concurrent checkouts can both pass the new validation; one may later fail inventory deduction, or the order accepted event may already have been published depending on event ordering/consumption.

**Recommendation:** keep the UI pre-check if desired, but do not rely on it for correctness. The authoritative reservation/deduction should be synchronous/transactional in the checkout flow or expose an atomic reservation API from Inventory that Basket.Api calls before accepting the checkout. At minimum, document that this is only a user-facing early validation and ensure downstream failures are handled and reflected in order state.

---

### 4. **High — new repository method hides upstream errors as “basket not found”**
**File:** `be/src/Project/Web/Seller/Seller.Web/Areas/Orders/Repositories/Baskets/BasketRepository.cs`

`GetBasketByIdAsync()` returns `default` for every non-success response:

```csharp
if (response.IsSuccessStatusCode && response.Data is not null)
{
    return response.Data;
}

return default;
```

Most other repository methods throw `CustomException(response.Message, statusCode)` on non-success. This method converts unrelated failures — `401`, `403`, `500`, service unavailable, malformed response — into `null`, which `BasketService` reports as `BasketNotFound`.

**Impact:** wrong status/message returned to the frontend, operational issues hidden, and possible authorization failures misreported.

**Recommendation:** follow existing repository conventions:

```csharp
if (response.IsSuccessStatusCode && response.Data is not null)
{
    return response.Data;
}

if (!response.IsSuccessStatusCode)
{
    throw new CustomException(response.Message, (int)response.StatusCode);
}

return default;
```

Then handle genuine empty/missing basket separately.

---

### 5. **Medium — potential `NullReferenceException` / `InvalidOperationException` paths in `BasketService`**
**File:** `be/src/Project/Web/Seller/Seller.Web/Areas/Orders/Services/Basket/BasketService.cs`

Several assumptions can produce a `500` instead of a controlled validation error:

```csharp
if (basket.Items.Any())
```

will throw if `Items` is `null`.

```csharp
stockItems.Select(x => x.ProductId.Value)
.GroupBy(x => x.ProductId.Value)
```

will throw if an item has no `ProductId`.

```csharp
inventoryProducts.GroupBy(...)
outletProducts.GroupBy(...)
```

will throw if repository returns `null` for a successful empty/no-content response.

**Impact:** invalid/corrupt basket data or unexpected downstream responses can break checkout with a generic error.

**Recommendation:** use existing `OrEmptyIfNull()` patterns and validate required fields explicitly:

```csharp
var items = basket.Items.OrEmptyIfNull().ToList();

if (items.Any(x => !x.ProductId.HasValue))
{
    throw new CustomException(_orderLocalizer.GetString("ProductsNotFound").Value, (int)HttpStatusCode.BadRequest);
}

var inventoryProducts = (await ...).OrEmptyIfNull();
```

---

### 6. **Medium — validation is duplicated and hard to maintain for stock vs outlet**
**File:** `be/src/Project/Web/Seller/Seller.Web/Areas/Orders/Services/Basket/BasketService.cs`

The stock and outlet branches are almost identical: filter items, fetch availability, group requested quantities, compare, throw formatted error. This duplication makes future fixes easy to apply to one branch but miss the other.

**Recommendation:** extract a shared helper for grouped quantity validation, or model stock/outlet as two validation strategies. This would also make unit tests easier to write.

---

### 7. **Low / UX — frontend assumes all failed checkout responses contain JSON with `message`**
**File:** `fe/projects/Seller.Portal/src/areas/Orders/components/OrderForm/OrderForm.js`

The PR changed failed checkout handling to:

```javascript
toast.error(jsonResponse.message);
```

That improves server-side validation visibility, but `response.json()` can still fail for empty, HTML, redirected, or non-standard error responses. The catch then shows the generic message, which is acceptable, but a safer pattern would fallback if `jsonResponse.message` is missing:

```javascript
toast.error(jsonResponse.message || props.generalErrorMessage);
```

---

### 8. **Low / Style — minor formatting and consistency issues**
**Files:**
- `BasketCheckoutApiController.cs`
- localization `.resx` files

Examples:

```csharp
private readonly IBasketService _basketService; 
```

has trailing whitespace.

The changed `.resx` files still show `\ No newline at end of file`. Not functionally important, but worth fixing to reduce diff noise and keep repository formatting consistent.

---

dawiddworak88 (changes requested):
Hi @PiotrZaranek, there is one more architectural thing that we'll need to change. At the moment we're creating a call to inventory from the basket service. I don't want to create a dependency between basket and inventory. It should be the BFF layer (Buyer.Web and / or Seller.Web) that first calls the inventory and then basket.

dawiddworak88 (changes requested):
We're almost there :) 
---

## 🔴 High — Correctness / Crash risks

**1. Possible `NullReferenceException` when the inventory/outlet API succeeds but returns no data.**
In `BasketCheckoutApiController.Checkout`:
```csharp
var inventories = inventoriesIds.Any()
    ? await _inventoryRepository.GetInventoryProductByProductIdsAsync(...)  // can return null
    : Enumerable.Empty<InventoryItem>();
```
`GetInventoryProductByProductIdsAsync` (and the outlet equivalent) `return default;` — i.e. **null** — on a success response with no payload. That null is then passed into `ValidateStockOutletQuantities`, which calls `inventoryItems.Any()` / `outletItems.Any()` directly and will throw an NRE.
*Fix:* guard with `.OrEmptyIfNull()` either at the repository return, in the controller, or at the top of `ValidateStockOutletQuantities` (the helper is already imported and used for `items`).

**2. `ProductId` null-check has the wrong scope and can reject valid checkouts.**
```csharp
if (items.Any(x => x.StockQuantity > 0 || x.OutletQuantity > 0))
{
    if (items.Any(x => x.ProductId.HasValue is false))   // checks ALL items
        throw ... ProductsNotFound;
```
The inner check looks at **every** basket item, not only the stock/outlet ones. If a basket legitimately contains a non-stock/custom line without a `ProductId`, but also has at least one stock item, the whole checkout is rejected with "Products not found." The predicate should be scoped to the items actually being validated, e.g. `items.Any(x => (x.StockQuantity > 0 || x.OutletQuantity > 0) && x.ProductId.HasValue is false)`.

---

## 🟡 Medium — Robustness / Consistency

**3. Inconsistent `.Value` usage on localized strings.**
Most `CustomException` throws use `_localizer.GetString("Key").Value`, but several new ones pass the `LocalizedString` directly:
- Controller: `GetString("ProductsNotFound")` (no `.Value`)
- `BasketService`: `GetString("InventoryNotFound")` and `GetString("OutletNotFound")` (no `.Value`)

It compiles via implicit `LocalizedString → string` conversion, but it's inconsistent and brittle. Make them all `.Value` for uniformity.

**4. Hidden precondition: `BasketService` assumes `ProductId` is non-null.**
`ValidateQuantities` does `items.GroupBy(x => x.ProductId.Value)`. This only works because the controller pre-validates `ProductId.HasValue`. If `BasketService` is ever called from another path, it throws on null. Consider validating/filtering inside the service so the component is self-contained, rather than relying on an external caller's guard.

**5. Redundant `BasketNotFound` validation.**
The controller already rejects a null/empty basket before calling the service, yet `ValidateStockOutletQuantities` repeats the same empty check and throws `BasketNotFound`. Harmless but duplicated logic in two layers — pick one owner for the rule.

**6. `GetBasketByIdAsync` control flow is convoluted.**
```csharp
if (response.IsSuccessStatusCode && response.Data is not null) return response.Data;
if (response.IsSuccessStatusCode is false) throw ...;
return default;
```
Reordering to "throw on failure first, then return data, then default" (as `InventoryRepository` does) is clearer and avoids the silent `null` success path that feeds back into bug #1.

SebaKas21 (changes requested):
## Follow-ups
- 🟡 **Culture consistency** — only `GetOutletByProductId` uses `CurrentUICulture`;
  the other actions in `OutletController` still use `CurrentCulture`. Worth unifying.
- 🟢 `GetAvailableProductsOutlets`: drop the now-ignored `.Include(...).AsSingleQuery()`
  (EF strips `Include` after `GroupBy`).

SebaKas21 (approved):
**Sequential round-trips.** `GetBasketById` → inventory → outlet are awaited
  in sequence on the checkout hot path; inventory/outlet are independent and
  could run via `Task.WhenAll`.

be/src/Project/Web/Seller/Seller.Web/Areas/Orders/ApiControllers/BasketCheckoutApiController.cs
```diff
@@ -51,6 +65,41 @@ public async Task<IActionResult> Checkout([FromBody] CheckoutBasketRequestModel
             var token = await HttpContext.GetTokenAsync(ApiExtensionsConstants.TokenName);
             var language = CultureInfo.CurrentUICulture.Name;
 
+            var basket = await _basketRepository.GetBasketByIdAsync(token, language, model.BasketId);
+
+            if (basket is null || basket.Items.Any() is false)
```
dawiddworak88: Let's use basket.Items.OrEmptyIfNull().Any() so that there is no exception when basket.Items is null

dawiddworak88 (approved):

PR #559, opened 2026-06-10
Fix catalog product indexing timeouts with missing indexes

SebaKas21:
@dawiddworak88 indexation queries ProductImages/ProductVideos/ProductFiles with the ProductId filter for each battch 

That's why timeouts were occurring again.


dawiddworak88 (changes requested):
Shouldn't it be the hotfix to main?

PR #560, opened 2026-06-10
Fix catalog product indexing timeouts with missing indexes

SebaKas21:


be/src/Project/Services/Catalog/Catalog.Api/Infrastructure/Migrations/20260611055401_AddProductIndexingIndexes.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Catalog.Api.Infrastructure.Migrations
+{
+    [DbContext(typeof(CatalogContext))]
+    [Migration("20260610120000_AddProductIndexingIndexes")]
+    public partial class AddProductIndexingIndexes : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateIndex(
```
dawiddworak88, line 13: It would be good Sebastian to generate the migrations with a command instead of AI. It looks as if .Designer file wasn't generated (like for other migrations)

dawiddworak88 (commented):
One small remark / question before we merge

dawiddworak88 (approved):

PR #561, opened 2026-06-10
Added numbers of mirrors and led price drivers

SebaKas21:


dawiddworak88 (changes requested):
Let's deliver it via hotfix to main

PR #570, opened 2026-06-18
Fix error when deleting product attribute and improve validation

PiotrZaranek:


SebaKas21 (changes requested):
### 🟡 Confirm before merge — semantic change for item deletion
Merging the two `AssertCategorySchemaReference` overloads changes the validation
for `DeleteProductAttributeItemAsync`:
- **Before:** blocked only if a category schema used *this item's specific value*
  (`$.definitions` → `anyOf` → `title IN (item translations)`).
- **After:** blocks if the *parent attribute* is referenced at all
  (`$ref = #/definitions/{ProductAttributeId}` in `$.properties`).

This is coarser — it may now block deleting an unused value just because its
attribute is referenced, and probes a different JSON branch than before. Same
error key (`ProductAttributeInUseByCategories`). Please confirm per-attribute
granularity is intended for item deletion (this is a behavior change, not a pure
refactor).

### 🟢 Minor
- `AssertCategorySchemaReference` returns `Task` but lacks the `Async` suffix.
- `$@"..."` no longer needs the `$` (no interpolation left).

dawiddworak88 (approved):
Two small improvement ideas and approval :) 

1. Drop the unnecessary `$` on the raw SQL__ (`$@"` → `@"`). The only value is already a `SqlParameter`; the interpolation prefix adds injection risk for future edits and forces `{`/`}` doubling. (Now my top item.)

2. Name the method `AssertCategorySchemaReferenceAsync` for convention consistency

SebaKas21 (approved):

PR #571, opened 2026-06-30
Change product attribute indexing

PiotrZaranek:


SebaKas21 (changes requested):
[P1] Regression: full-text search by attributes lost
  File:  CompositionRoot.cs:42 + ProductSearchRepository.cs:62-70
  What:  FormData changed to Index(false), and query_string does not search
         nested fields. New ProductAttributes.ValueKeywords/ValueText are NOT
         added to the query => products can't be found by attribute value.
  Fix:   Add a nested query on ProductAttributes.ValueKeywords in GetAsync,
         or consciously confirm the loss of this feature.

[P2] Indexing crash for object-typed values
  File:  BulkProductIndexingRepository.cs:373 (and cast at :320)
  What:  token.Value<string>() on a JObject throws InvalidCastException
         (resolves to IEnumerable<JToken>.Value<string>() extension). Old code
         used Convert.ToString() and didn't crash. The (JObject)schemaProperty.Value
         cast now runs for EVERY schema property.
  Fix:   Use token.ToString()/Convert.ToString; guard the schema cast
         (check schemaProperty.Value is JObject).

[P2] Performance: ResolveEnumTitle (recursive JSONPath)
  File:  BulkProductIndexingRepository.cs:400
  What:  $.definitions..anyOf[?(@.enum[0]=='id')].title run per attribute value
         * product * language. Expensive and repeated.
  Fix:   Build an id->title dictionary once per schema (when building schemaCaches).

[P3] Culture-dependent number format + dead branch
  File:  ProductsService.cs:480
  What:  ValueNumeric.ToString() without InvariantCulture; branch is dead anyway
         (ValueKeywords always set for numbers).
  Fix:   InvariantCulture, or remove the unreachable branch.

[P4] Cleanliness / data duplication
  - Redundant ValueBoolean/ValueNumeric vs ValueKeywords (index size).
  - Broken indentation at ProductsService.cs:463-468 (cosmetic).

SebaKas21 (changes requested):
[P1] ValueNumeric never set
     File: BulkProductIndexingRepository.cs  (Integer/Float case)
     Numbers stored only as keyword strings; mapped double field is always null
     => no range filter / sort / numeric aggregation. Breaks the aggregation goal.
     Fix: attribute.ValueNumeric = token.Value<double>();

[P2] Nested Prefix on ValueKeywords is case-sensitive
     File: ProductSearchRepository.cs  (GetAsync searchTerm block)
     'red' won't match stored 'Red'; also prefixes true/false + labels.
     Fix: keyword normalizer (lowercase) or match a text sub-field.

SebaKas21 (approved):

PR #573, opened 2026-07-07
Default sorting

SebaKas21:


dawiddworak88 (changes requested):
Hi @SebaKas21, here a few remarks that might be relevant in order to assure the same items are returned from various endpoints upon pagination. I believe Sonnet 5 might be helpful here since I generated the plan with Ops 4.8. Would you have a look at them?

**Proposed solution:**

### 1. `Foundation.Extensions/ExtensionMethods/IQueryableExtensions.cs` (SQL/EF)
- Add `private const string TieBreakerPropertyName = "Id";`
- Add helper `EnsureStableTieBreaker<T>(string orderByString)` that appends `Id ascending` (Dynamic LINQ syntax) unless the type has no `Id` property, or `Id` is already in the sort.
- Call it right before `source.OrderBy(orderByString)`.
- Result: every SQL sort (empty → `CreatedDate desc, Id asc`; explicit → `<user sort>, Id asc`) is deterministic.

### 2. `Foundation.Search/Extensions/StringExtensions.cs` (Elasticsearch)
- Fix `.keyword` mapping: unwrap nullables and only append `.keyword` for `string` (via a `ResolveElasticField` helper). This makes `DateTime?`, `Guid`, `int`, `bool` sort correctly.
- Add `private const string ElasticTieBreakerPropertyName = "ProductId";`
- After the loop, append `productId ascending` unless already sorted by it and only if the type has that property.

### 3. `Foundation.Catalog/DependencyInjection/CompositionRoot.cs`
- Verify/ensure `productId` is sortable. `AutoMap()` maps `Guid` to keyword by default, so this should already work. I'll confirm during Act and only add explicit mapping if needed.

### 4. Tests — `Giuru.UnitTests`
- Extend `IQueryableExtensionsTests.cs`:
  - duplicate `CreatedDate` → deterministic order via `Id`
  - explicit sort with duplicate key → `Id` tie-breaks
  - explicit sort by `Id` → not duplicated
  - type without `Id` → no throw, no tie-breaker
  - simulated page1/page2 over tied data → no overlap, full coverage
- Add `StringExtensionsTests.cs`:
  - `string` → `.keyword`; `DateTime`/`DateTime?`/`Guid`/`int` → no `.keyword`
  - empty orderBy on type with `CreatedDate` → `createdDate desc` + `productId asc`
  - no duplicate tie-breaker when caller already sorts by `ProductId`
- Update `Giuru.UnitTests.csproj` to add a `ProjectReference` to `Foundation.Search.csproj`.

dawiddworak88 (approved):
🎖️ I'd appreciate checking on testing env  especially from the side of admin and API after merge. Thanks @SebaKas21 !

